PR #1, opened 2022-12-28
remove unused var, small improvements

Disservin:
1. remove unused imports and vars
2. use "bit count" for material, 3-4x faster than old counting method
3. take all pgns that are in the `./pgn` folder

### DIFF
--- a/scoreWLDstat.py
+++ b/scoreWLDstat.py
@@ -3,13 +3,9 @@ from collections import Counter
 import chess
 import chess.pgn
 import re
-import sys
-import random
 import json
 import os
 import argparse
-import io
-
 
 class PosAnalyser:
     def __init__(self, plies):
@@ -45,7 +41,6 @@ class PosAnalyser:
             # look at the game,
             plies = 0
             board = game.board()
-            fen = None
             for node in game.mainline():
 
                 plies = plies + 1
@@ -58,7 +53,6 @@ class PosAnalyser:
                 m = p.search(node.comment)
                 if m:
                     score = m.group(1)
-                    depth = int(m.group(2))
                     m = mateRe.search(score)
                     if m:
                         if m.group(1) == "+":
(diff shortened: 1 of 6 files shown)


Disservin:
With 140MB of pgns the combination step took 4s.

vondele:
1. and 2. look good to me, and I could merge them. However 3 doesn't work for me. Reading all pgns in memory only works for smaller sets of pgns. I have used the tools on more than 100GB of data. The parallelism is over files, typical collection of pgns I would use is like 100k files.

Disservin:
Yeah okay in that case it would be a slowdown, curious to know is there really a need for that many pgns? What’s like the NormalizePawnValue diff between 1GB and 100GB?

vondele:
To just extract the NormalizePawnValue value, I think far fewer pgns will do. I got very reasonable results with just 200k games, even few could probably be enough. A lot of games is mostly useful to get the full 2D contour plot look good.

Disservin:
Just read 3. doesnt work for you ? `take all pgns that are in the ./pgn folder` why exactly does this not work for you?

vondele:
This: `take all pgns that are in the ./pgn folder` will work. I was referring to loading all in memory.

Disservin:
Ah okay. Then please have a look at the changes code ^^

scoreWLDstat.py
```diff
@@ -101,43 +86,37 @@ def ana_pos(self, filename):
...
+    parser.add_argument(
+        "--matching_plies",
+        type=int,
+        default=6,
+        help="Number of plies that the material situation needs to be on the board (unused).",
     )
+    args = parser.parse_args()
+
+    # use all files in ./pgns that end with pgn
+    path = "./pgns"
```
vondele, line 101: so, what about making that actually an argument of the script (the default can be ./pgns) ?

vondele:
Thanks, works for me!

PR #3, opened 2023-01-10
display progress

Disservin:
`pip install tqdm`

it/s may not 100% reflect how long each pgn file took


Disservin:
image

how it looks like

PR #6, opened 2023-07-11
cpp implementation of scoreWLDstat.py

Disservin:


vondele:
I think this should also remove the python version, and update the readme. The ~7x speedup is nice to have.

Disservin:
Now down to 71 seconds for an entire 36GB dataset.
Due to improved splitting of the string when we are extracting the score and using a faster implementation of `std::stof`

```
Found 48134 pgn files, creating 128 chunks for processing.
Progress: 128/128
Time taken: 71s
Retained 1481340849 scored positions for analysis.
```

PR #7, opened 2023-08-20
suptitle

robertnurnberg:
Change the `suptitle` (could also just write "WDL"?) and applied black.

Btw, what is the easiest way to obtain the latest file `scoreWLDstat.json`? If I had it, I could update the `pgn` as well.

### DIFF
--- a/scoreWDLana_moves_fit.py
+++ b/scoreWDLana_moves_fit.py
@@ -36,7 +36,7 @@ save_dpi = 300
 fig, axs = plt.subplots(
     2, 3, figsize=(11.69 * 1.5, 8.27 * 1.5), constrained_layout=True
 )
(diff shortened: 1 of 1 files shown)


vondele:
no easy way to have the json, it is quite large.

vondele:
I've made a copy of data available, for testing, basically current master https://drive.google.com/file/d/1ckSG8g37u9b5ZONoBwG5virYlrb0kFE6/view?usp=sharing

robertnurnberg:
Thanks. Will download and test later.

robertnurnberg:
Using the data I get this plot, which differs from master. (Check polynomial at bottom left.)

Is the fitting deterministic?
WDL_model_summary

vondele:
yeah, it is a different data set (i.e. August data).

PR #8, opened 2023-08-20
readme

robertnurnberg:
Draft to allow further discussions:

1. <s>For the first bullet point, do we have a command line or a script we can upload to show how to download games from fishtest?</s> (added vondele's download script)
2. <s>Is `main` a good name for the executable? Alternatives would be simply `scoreWDLstat`, or `pgnscraper` or `collectWDLdata`.</s> (Joost has spoken)
3. <s>Is the Windows detection in the Makefile needed at all? Also the `NATIVE` switch change in case of `Darwin`? If not, I would delete both.</s> (discuss elsewhere)
4. <s>I changed the filename of the python script `scoreWDLana_moves_fit.py` based on vondele's change to the readme in https://github.com/official-stockfish/WDL_model/commit/c027b69d4a1a8acd965eafacebd5ee7e457ffcaa. Is that OK? The names of the other two (and the description of what they do at the bottom of the readme) could also be changed.</s> (OK as per Joost)
5. <s>Is the second entry in the json tuples move number or ply number? From the code it looks like move number (as stated in readme), but the comment `@brief [WLD, plies, matcount, score]` in the .cpp file mentions plies.</s> (move number seems correct, update comment in .cpp separately)

Another question: should we rename the `master` branch to `main`? (I'd be in favour.)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,4 +3,5 @@
 *.pgn
 
 pgns*
-main*
+scoreWDLstat
+scoreWDLstat.exe
--- a/Readme.md
+++ b/Readme.md
@@ -1,15 +1,39 @@
 # Generate SF WDL model based on data
 
+Stockfish's "centipawn" evaluation is decoupled from the classical value
+of a pawn, and is calibrated such that an advantage of
+"100 centipawns" means the engine has a 50% probability to win
+from this position in selfplay at fishtest LTC time control.
+
 ## Usage
 
-To generate the WDL model three steps are needed:
+To update Stockfish's internal WDL model, the following steps are needed:
 
-1. You first need to compile `scoreWDL.cpp`. The simplest way to do that is to use the supplied Makefile, in which case you only need to run `make` in the current directory. This will output an executable named `main`.
+1. Obtain a large collection of engine-vs-engine games 
+(at fishtest LTC time control) in pgn format and
+save the pgn files in the `pgns` folder. This can, for example, be achieved
+by running `python download_fishtest_pgns.py --path pgns` once a day.
 
-2. Run `main` to parse a large collection of pgn files providing the data. By default the script will look for pgn files in the `pgns` folder that end with `.pgn`, unless a different directory is specified using `main --dir \<path-to-dir\>`.
+2. Use `make` to compile `scoreWDLstat.cpp`, which will produce an executable
+named `scoreWDLstat`.
 
-3. Run scoreWDL.py to compute the model parameters
-   For this step, specify the with option `--NormalizeToPawnValue` the correct number to convert pawn scores (as assumed to be in the pgn) to the units internally used by the engine.
+3. Run `scoreWDLstat` to parse the pgn files in the `pgns` folder. A different
+directory can be specified with `scoreWDLstat --dir <path-to-dir>`. The
+computed WDL statistics will be stored in a file called `scoreWDLstat.json`.
+The file will have entries of the form `"('D', 1, 78, 35)": 668132`, meaning
+this tuple for `(outcome, move, material, eval)` was seen a total of 668132
+times in the processed pgn files.
+
+4. Run `python scoreWDL.py` to compute the WDL model parameters from the
+data stored in `scoreWDLstat.json`. The script needs as input the value
+`--NormalizeToPawnValue` from within Stockfish's
+[`uci.h`](https://github.com/official-stockfish/Stockfish/blob/master/src/uci.h),
+to be able to correctly convert the centipawn values from the pgn files to
+the unit internally used by the engine. The script will output the new
+values for `NormalizeToPawnValue` in
+[`uci.h`](https://github.com/official-stockfish/Stockfish/blob/master/src/uci.h)
+and `as[]`, `bs[]` in
+[`uci.cpp`](https://github.com/official-stockfish/Stockfish/blob/master/src/uci.cpp). See e.g. https://github.com/official-stockfish/Stockfish/pull/4373
 
 ## Results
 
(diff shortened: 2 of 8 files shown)


Disservin:
3. i thought the march native was beneficial but on earlier Mac OS compilers this is not supported, but the switch could be simplified.
4. Also applied to 2. I personally find it annoying that (almost) all files start with scoreWDL. This is annoying when you use tab completion.
5. in the cpp code it’s actually plies, didn’t check the python part but I’m pretty sure it was plies in the old version as well. 

Btw I think the file name is not correct anymore in step 4 of the readme

robertnurnberg:
Thanks for your feedback.

> 3. i thought the march native was beneficial but on earlier Mac OS compilers this is not supported, but the switch could be simplified.

Is this for g++? If so, could you send me a link? IMHO the Makefile should be as simple as possible, and having only g++ would be fine. If `-march=native` is really an issue on some OS, we could deal with that switch as in SF. I.e. invoked from outside the Makefile.

> 4. Also applies to 2. I personally find it annoying that (almost) all files start with scoreWDL. It’s annoying when you use tab completion.

Ok, I wait for Joost to maybe suggest new names. `main` is not very fitting though, as that program merely collects the json data from the pgns. Could use `pgns2json`, `pgnanalysis` or `pgnscraper` instead?

> 5. in the cpp code it’s actually plies, didn’t check the python part but I’m pretty sure it was plies in the old version as well.

Would be quite bad if cpp code did things differently from python code. ;) The way I understand the lines `plies++;` and `const int plieskey = (plies + 1) / 2;` the latter will be like `1, 1, 2, 2, 3, 3` etc, so moves.

> Btw I think the file name is not correct anymore in step 4 of the readme

The change to 4. in the readme comes from https://github.com/official-stockfish/WDL_model/commit/c027b69d4a1a8acd965eafacebd5ee7e457ffcaa, and this PR reflects that change on the filesystem level.

Disservin:
3. Never used g++ on mac. Apple's clang fails with `-march=native` if you version is < 17, but the Makefile isnt very large so I dont really see this as a problem.
4. `pgns2json` implies that it converts a pgn into a json. The other ones are more fitting, or maybe something wit h`collect`... ?
5. I only had a look from my phone yesterday evening but if im doing `const int plieskey = (plies + 1) / 2;`  then all is good and the variable & comment can be renamed.

I know, I just proposed that you fix the name with this pr as well.

robertnurnberg:
Ok, best to leave the Makefile as it is for now. Can always come back to that.

So this PR should only touch the readme and then make the filenames in the repo etc consistent with it. I will await suggestions for names from Joost and then update the PR accordingly.

Readme.md
```diff
@@ -2,31 +2,41 @@
...
+named `main`.
+
+3. Run `main` to parse the pgn files in the `pgns` folder. A different directory
+can be specified with `main --dir <path-to-dir>`. The computed WDL statistics
+will be stored in a file called `scoreWLDstat.json`. The file will have
+entries of the form `"('D', 1, 78, 35)": 668132", meaning this tuple for
+`(outcome, move, material, eval)` was seen a total of 668132 times in the
+processed pgn files.
+
+4. Run `python scoreWDL.py` to compute the WDL model parameters from the
```
Disservin, line 20: ``python scoreWDL.py`` if you are already editing the readme, I would simply correct this too.

robertnurnberg: You mean going back to the old name `scoreWDLana_moves_fit.py` ? Will do if @vondele confirms that is the name he wants for that script.

Disservin: Going back ? I think it was simply a mistake by vondele. 
`Run python scoreWDL.py` makes no sense currently because such file doesnt even exist?

robertnurnberg: I believe this PR makes files and readme consistent. (It renames the offending python script.) Maybe not in the way vondele intended, granted. We can discuss the best names for all scripts and the .cpp file.

Disservin: resolved, i missed that the pr also fixes the name of the file.

robertnurnberg:
I have now added an edited version of the download script from vondele's discord link, and mention it in bullet point 1.

Readme.md
```diff
@@ -2,31 +2,44 @@
...
-3. Run scoreWDL.py to compute the model parameters
-   For this step, specify the with option `--NormalizeToPawnValue` the correct number to convert pawn scores (as assumed to be in the pgn) to the units internally used by the engine.
+To update Stockfish's internal WDL model, the following steps are needed:
+
+1. Obtain a large collection of engine-vs-engine games in pgn format and
+save the pgn files in the `pgns` folder. This can, for example, be achieved
+by running `python download_fishtest_pgns.py --path pgns` once a day.
+
+2. Use `make` to compile `scoreWDLstat.cpp`, which will produce an executable
+named `main`.
```
vondele, line 12: what about using the renaming to change the executable into `scoreWDLstat` instead of main, that's a bit more common (needs Makefile adjustment).

Readme.md
```diff
@@ -9,9 +9,10 @@ save the pgn files in the `pgns` folder. This can, for example, be achieved
 by running `python download_fishtest_pgns.py --path pgns` once a day.
 
 2. Use `make` to compile `scoreWDLstat.cpp`, which will produce an executable
-named `main`.
+named `scoreWDLstat`.
 
-3. Run `main` to parse the pgn files in the `pgns` folder. A different directory
+3. Run `scoreWDLstat` to parse the pgn files in the `pgns` folder. A different 
+directory
 can be specified with `main --dir <path-to-dir>`. The computed WDL statistics
```
vondele, line 16: executable name

download_fishtest_pgns.py
```diff
@@ -0,0 +1,73 @@
+import urllib.request, urllib.error, urllib.parse
+import argparse, time, re, os
+
+parser = argparse.ArgumentParser()
+
+parser.add_argument(
+    "--path",
+    default="./",
```
Disservin, line 8: can we change this to `"./pgn"` so that the files are directly stored in the default directory for the cpp part ?
I'm not sure if the directory will be automatically created if it is missing though.

vondele: that would need some changes on my end as well.
I propose we wait a bit with that.
In the long run, I think it would be useful to bring some more order in this, e.g. download pgns per day, or per test, and allow for a collection of directories to be analyzed.
This would be allow us to e.g. track the evolution of the normalization factor over time more easily.
This needs some more work though in the scripts.

robertnurnberg: Default behaviour is as in the original script. See readme for usage as you want it. Directory must exist.

Readme.md
```diff
@@ -2,31 +2,44 @@
...
-  produces scoreWDLstat.json : extracted data from games
-  "('D', 1, 78, 35)": 668132, -> outcome = draw, move = 1, material = 78, eval = 35 cp -> 668132 positions
-- scoreWDLana_moves_fit.py : fit some models
+Other scripts that can be used to visualize different WDL data:
 
-- scoreWDLana_moves.py : similar to above, analyze wrt to moves
-- scoreWDLana_material.py : similar to above analyze wrt to material
+- `scoreWDLana_moves.py` : similar to `scoreWDL.py`, analyze wrt to moves
+- `scoreWDLana_material.py` : similar to `scoreWDL.py`, analyze wrt to material
+---
```
robertnurnberg, line 44: Are these two script names (and the description of what they do) OK? @vondele 

I was not sure about the latter....

vondele: yes, that's OK.

robertnurnberg:
I think this is ready for merge. My latest commit added a paragraph in the readme, fixed some more WLD left-overs and edited `.gitignore`.

.gitignore
```diff
@@ -3,4 +3,4 @@
 *.pgn
 
 pgns*
-main*
+scoreWDLstat
```
Disservin, line 6: shouldnt this be `scoreWDLstat*` to also match `scoreWDLstat.exe` on windows ?

robertnurnberg: I have added .exe explicitly for now

PR #9, opened 2023-08-22
small code edits using OOP

robertnurnberg:
This is a small rewrite of @Disservin's cpp code. I tried to use a bit more OOP, and it is now a bit easier for me to read.

Could you check if you see any slowdown? It should be functionally equivalent, but better be safe than sorry.

Apart from collecting all the tuple data and member functions within `struct Key`, I also avoid passing by value/returning of large objects and instead went for passing by reference.

### DIFF
--- a/scoreWDLstat.cpp
+++ b/scoreWDLstat.cpp
@@ -1,4 +1,3 @@
-
 #include <atomic>
 #include <chrono>
 #include <cmath>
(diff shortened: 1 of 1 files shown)


Disservin:
Analysis looks good.
```
oop:
Found 46353 pgn files, creating 128 chunks for processing.
Progress: 128/128
Time taken: 80s
Retained 1424144030 scored positions for analysis.

master:
Found 46353 pgn files, creating 128 chunks for processing.
Progress: 128/128
Time taken: 85s
Retained 1424144030 scored positions for analysis.
```

scoreWDLstat.cpp
```diff
@@ -18,36 +17,46 @@ namespace fs = std::filesystem;
...
 };
 
-/// @brief [WLD, plies, matcount, score] tuple key for the position map
-using map_key_t = std::tuple<Result, int, int, int>;
-
-struct key_hash {
-    std::size_t operator()(const map_key_t &k) const {
+struct Key {
+    Result outcome;
+    int move, material, score;
```
Disservin, line 29: can we rename `move` to `move_count`  or `move_number` or something similar? 
To not confuse it with an actual chess move.

robertnurnberg:
Nice, thank you. I wonder if the small speedup is a fluke or down to us now avoiding these temporary copies of `map_t` objects when returning results from `ana_files`. As these objects could be quite large, maybe the speedup is real.

Disservin (changes requested):

Disservin:
> Nice, thank you. I wonder if the small speedup is a fluke or down to us now avoiding these temporary copies of `map_t` objects when returning results from `ana_files`. As these objects could be quite large, maybe the speedup is real.

Mh just a slight fluke it seems but it sure avoids some unnecessary stuff now

scoreWDLstat.cpp
robertnurnberg: Hm, I like the shortness of `move`, and I followed the convention from the readme: `outcome = draw, move = 1, material = 78, eval = 35 cp`. And as these are properties of `Key`, I hope no confusion can arise.

In any case, I would wait for feedback from @vondele . If we go for different name, then maybe `moveNr` (and also consistent in code comments and Readme?).

vondele: it is move also in the python scripts, I'd rather keep it like that for now, can be changed in a later PR to moveNr or similar

Disservin (approved):

PR #11, opened 2023-08-22
add clangformat

Disservin:


vondele:
Do you want to make a 'make format' Makefile target for this ? (and a make clean while at it?).

Makefile
```diff
@@ -20,3 +20,7 @@ endif
 
 all:
 	$(CXX) $(CXXFLAGS) $(NATIVE) -o scoreWDLstat scoreWDLstat.cpp
+format:
+	clang-format -i scoreWDLstat.cpp
+clean:
+	rm scoreWDLstat.json
```
vondele, line 26: I meant removing the executable not the output result. Usually, I would not remove the .json.

Makefile
```diff
@@ -23,4 +23,4 @@ all:
 format:
 	clang-format -i scoreWDLstat.cpp
 clean:
-	rm scoreWDLstat.json
\ No newline at end of file
+	rm scoreWDLstat scoreWDLstat.exe
```
vondele, line 26: `rm -f scoreWDLstat scoreWDLstat.exe`

PR #16, opened 2023-08-25
Add python dependencies install to readme

dede1751:


vondele:
what about adding a requirements.txt file ?

dede1751:
I did it this way to keep the same format as pytorch-nnue, I'll happily do the requirements file

dede1751:
Committed the versions pip defaults to on Python 3.11, as I do not know which Python version is considered default for this project.

vondele:
Admittedly, no idea if it is better to specify a specific version or leave the version number unspecified for these dependencies. I don't think we use anything that is really specific in functionality so my approach would have been to leave it unspecified. I'm happy to adopt was is considered best practice however.

robertnurnberg:
I'd think our python scripts also work for older versions of these packages, so would suggest to not state version numbers. Should we also add `numpy`?

dede1751:
Numpy is already a Scipy dependency, so that shouldn't be necessary. I admit I'm not too familiar with Python conventions and it seemed preferred to put strict checks in the requirements, though I personally agree leaving it up to whatever pip does is reasonable.

Disservin:
Mh I have no strong feelings about the version thing, but if anything were to ever break because of either package then potential users at least know that it worked with version xx ..., but I think the chances for that are fairly small ; )

Also maybe perhaps specify that the steps for the venv are only for POSIX. Windows has a slightly different [syntax](https://docs.python.org/3/library/venv.html#how-venvs-work).

robertnurnberg:
I think _if_ we include version numbers, we should but the lowest numbers that still work (or close to it). Might be easiest to just not put any.
Regarding the virtualenv, these are mainstream packages, so once again might be simpler to just state `pip install -r requirements.txt` for the instructions, without any env.

.gitignore
```diff
@@ -5,3 +5,4 @@
 pgns*
 scoreWDLstat
 scoreWDLstat.exe
+env
```
robertnurnberg, line 8: Can now be removed

PR #18, opened 2023-08-28
add --moveTarget option

robertnurnberg:
This PR adds the option `--moveTarget` and some more help output, saves the graphical output for all three python scripts always to file, and applies some simplifications and cosmetic changes.

For ease of reviewing, this PR is functionally equivalent to master.

Eventually we could try to merge all three scripts into one, and have a very simple `main` that calls other functions/classes. Leave that for a follow-up for now.

### DIFF
--- a/scoreWDL.py
+++ b/scoreWDL.py
@@ -67,7 +67,7 @@ xs, ys, zwins, zdraws, zlosses = [], [], [], [], []
 for x, y in coords:
     xs.append(x)
     ys.append(y)
-    total = float(win[x, y] + draw[x, y] + loss[x, y])
+    total = win[x, y] + draw[x, y] + loss[x, y]
     zwins.append(win[x, y] / total)
     zdraws.append(draw[x, y] / total)
     zlosses.append(loss[x, y] / total)
--- a/scoreWDL.py
+++ b/scoreWDL.py
@@ -1,103 +1,76 @@
-import json, argparse, math, numpy as np, matplotlib.pyplot as plt
+import json, argparse, numpy as np, matplotlib.pyplot as plt
 from collections import Counter
 from ast import literal_eval
 from scipy.interpolate import griddata
 from scipy.optimize import curve_fit
 
-parser = argparse.ArgumentParser()
-
+parser = argparse.ArgumentParser(
+    description="Fit Stockfish's WDL model to fishtest game statistics.",
+    formatter_class=argparse.ArgumentDefaultsHelpFormatter,
+)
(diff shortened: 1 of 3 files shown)


scoreWDLana_material.py
```diff
@@ -3,59 +3,62 @@
...
-loss = Counter()
-draw = Counter()
-for k, v in inpdict.items():
-    (result, move, material, score) = k
-    if score < -400 or score > 400:
+win, draw, loss = Counter(), Counter(), Counter()
+# filter out (score, material) WDL data (i.e. move summed out)
+for (result, move, material, score), v in inpdict.items():
+    # exclude large scores and the endings of very long games
+    if abs(score) > 400:  #  or move < 0 or move > 120:
```
robertnurnberg, line 34: With the filter on move, this would become equivalent to `scoreWDL.py`. So include same move filter here? @vondele

scoreWDL.py
```diff
@@ -1,103 +1,76 @@
...
+# create (score, move) -> WDL ratio data
+coords = sorted(set(list(win.keys()) + list(draw.keys()) + list(loss.keys())))
+xs, ys, zwins, zdraws, zlosses = [], [], [], [], []
+for x, y in coords:
     xs.append(x)
     ys.append(y)
-    zs.append(win[coord] / total)
-    zdraws.append(draw[coord] / total)
-    zlosses.append(loss[coord] / total)
+    total = float(win[x, y] + draw[x, y] + loss[x, y])
```
robertnurnberg, line 70: Why `float` here? For speed? If there is no particular reason, could remove `float` here (and in the other two scripts). @vondele

vondele:
I think this looks good to me, I definitely agree that these 3 scripts could be merged into one. Two of them are mostly leftovers, never cleaned up properly. In the old days, having the wdl model depend on material didn't work as well as ply, but now this might have changed (and conceptually having it depend on material would be somewhat more natural). If the other scripts can be merged, even changing a little bit their functionality, that would be good.

scoreWDL.py
vondele: possibly for no good reason, at least I can't recall.

scoreWDLana_material.py
vondele: yes, see main comment below.

robertnurnberg:
I acted on Joost's feedback. May be ready to merge. 

Next up will be bigger refactoring, yielding only a single script that caters for all eventualities.

vondele (approved):

PR #19, opened 2023-08-28
unify python scripts

robertnurnberg:
This PR combines all the fitting and plotting into a single python script.

May need some testing from @vondele .

Adding fitting the WDL model to (score, material) from here may be possible, but I would rather leave that to vondele.


vondele:
so, my first question would be, does the new unified script called with all defaults generate the same results as the previous script?

robertnurnberg:
I think very nearly. The only functional difference is the filtering of moves < 3 at an earlier stage, and allowing 120 in the fitting for loop.

vondele:
Unless you have a particular reason to change these values, I would rather keep those identical, i.e. so we can more easily test things are the same.

Disservin:
I only flew over the code and from my perspective it now looks like everything is squashed togetherr.
If we are already unifying them I would expect the logic to be grouped a bit more into functions/classes.

robertnurnberg:
@vondele : latest push makes PR functionally equivalent to master. locally I confirm that the produced .png is the same

Once equivalence is confirmed, however, it may make sense to go back to the original form of this PR. You could comment on each `TODO` in sources to indicate whether you want master behaviour, or original PR.

@Disservin one thing at a time ;) I thought reviewing fewer changes might be easier. once the three scripts are merged, code could be refactored without functional changes.

vondele:
yes, I agree, refactoring can be done after the merge.

scoreWDL.py
```diff
@@ -120,193 +163,205 @@ def tick_function(X):
...
+    # convert to model, fit the winmodel a and b,
+    # for a given value of the move counter
+    #
+    scores, moves, winrate, drawrate, lossrate = xs, ys, zwins, zdraws, zlosses
+
+    model_ms, model_as, model_bs = [], [], []
+
+    grouping = 1
+    for m in range(args.moveMin, args.moveMax, grouping):
+    ### for m in range(args.moveMin, args.moveMax + 1, grouping): TODO
```
vondele, line 177: In the future, including 120 in the fit is good, i.e. that's what happens when the fit is used in SF.

scoreWDL.py
```diff
@@ -34,28 +58,47 @@
...
 win, draw, loss = Counter(), Counter(), Counter()
-# filter out (score, move) WDL data (i.e. material summed out)
-for (result, move, _, score), v in inpdict.items():
-    # exclude large scores and the endings of very long games
-    if abs(score) > 400 or move < 0 or move > 120:
+# filter out (score, yData) WDL data (i.e. material or move summed out)
+for (result, move, material, score), v in inpdict.items():
+    # exclude large scores and unwanted move numbers
+    if abs(score) > 400 or move < 0 or move > args.moveMax:
+    ### if abs(score) > 400 or move < 0args.moveMin or move > args.moveMax: TODO
```
vondele, line 88: here I would keep 0, i.e. better to be good 'out-of-the-book'. I assume I tested larger values, but after some testing decided for 0 (since the condition is a bit non-nonsensical with zero).

scoreWDL.py
```diff
@@ -26,6 +26,30 @@
...
+    default=120,
+    help="Maximum move number to consider for analysis.",
+)
+parser.add_argument(
+    "--yData",
+    choices=["move", "material"],
+    default="move",
+    help="Select y-axis data used for plotting and fitting.",
+)
+parser.add_argument(
```
vondele, line 47: in the future that could go, one can just always fit IMO.

scoreWDL.py
```diff
@@ -34,28 +58,47 @@
 )
 args = parser.parse_args()
 
+if args.fit:
+    figRows, figCols = 2, 3
+    title = "Summary of win-draw-loss model analysis"
+    pgnName = "WDL_model_summary.png"
+    assert args.yData == "move", f"--yData {args.yData} not yet implemented."  #  TODO
```
vondele, line 65: in the future this could go, i.e. fit always also for material.

scoreWDL.py
```diff
@@ -120,193 +163,205 @@ def tick_function(X):
...
+        axs[i, j].set_ylabel(ylabelStr)
 
-# for wins
+# for wins, plot between -1 and 3 pawns, using a 30x22 grid
 xmin = -((1 * args.NormalizeToPawnValue) // 100 + 1) * 100
 xmax = ((3 * args.NormalizeToPawnValue) // 100 + 1) * 100
-grid_x, grid_y = np.mgrid[xmin:xmax:30j, 10:120:22j]
+if args.yData == "move":
+    grid_x, grid_y = np.mgrid[xmin:xmax:30j, 10 : args.moveMax : 22j]
+    ### grid_x, grid_y = np.mgrid[xmin:xmax:30j, args.moveMin : args.moveMax : 22j] TODO
```
vondele, line 307: I guess I understand now your args.moveMin to 3. I think the reason for doing this 10 is that there is little data for the contour plot otherwise, for the fit this matters less I think, but showing the data looks a little ugly.

vondele:
So, IMO, this can be merged in this form, and improved in later iterations.

robertnurnberg:
Looking at your comments, best to remove `args.moveMin` completely? and hard-code all the values as in master?

vondele:
I think it can be left with a default of 0, it will allow for experimenting, but possibly the graphs that had 10 as a minimum just with hard-coded 10. Maybe we come up with a better way of visualizing this later on.

robertnurnberg:
Ok, I will do this later today. I agree that plotting only parts is not ideal. Rather show the ugly bit and be ready to explain it if it's not impactful.

Regarding `args.moveMin`, in future this may need to be replaced/shadowed by a `args.yDataMin` to allow specifying data ranges for fits based on material, and same for Max.

PR #22, opened 2023-08-29
allow fitting based on material

robertnurnberg:
This PR allows fitting based on material. (Based on my understanding of the code.)

No functional change for running `python scoreWDL.py` compared to https://github.com/official-stockfish/WDL_model/pull/21.

Attached sample output for the json data from https://github.com/official-stockfish/WDL_model/pull/7#issuecomment-1685286831 with the command 
`python scoreWDL.py scoreWLDstat.json --yData material --yDataMin 5 --yDataMax 78 --yDataTarget 68`.

PS: Of course, the C++ output for SF at this stage makes no sense, as SF is based on move.
WDL_model_summary

### DIFF
--- a/scoreWDL.py
+++ b/scoreWDL.py
@@ -21,28 +21,40 @@ parser.add_argument(
     help="Value needed for converting the games' cp scores to the SF's internal score.",
 )
 parser.add_argument(
-    "--moveTarget",
-    type=int,
-    default=32,
-    help="Move at which new rescaled 100cp should correspond to 50:50 winning chances.",
-)
-parser.add_argument(
     "--moveMin",
     type=int,
-    default=3,
-    help="Minimum move number to consider for analysis.",
+    default=0,
+    help="Lower move number limit for filter applied to json data.",
 )
 parser.add_argument(
     "--moveMax",
     type=int,
     default=120,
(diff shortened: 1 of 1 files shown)


robertnurnberg:
Agreed. But the model contour plots compared to the data look very wrong, not sure if I missed something in the maths behind the code. (I blindly used the same fitting functions and polynomials etc. as for moves)

The conflict was probably the `mom` change line 192, should be fixed after my forced push.

vondele:
It is potentially correct, the fits of the as and bs is probably not quite able to fit the low material part of the data, which propagates into the higher material part of the data. Restricting the fit to material > 10 or even 15 could work (or maybe providing much more data).

vondele:
I think the other interesting bit is that the cross-section at material 68 shows much better win rates than the model for eval near 0.

robertnurnberg:
I picked 68 completely at random (thinking was 1 minor piece and 2 paws exchanged on each side). :)

vondele:
so, with a little larger `--yDataMin 10` the data and  the model contour plots look pretty reasonable.
WDL_model_summary
This looks OK, and the code already looks much cleaner now.
Thanks!

PR #24, opened 2023-08-30
provide script to merge JSON files

robertnurnberg:
The json files use far less space than the full pgn collections, and grow only little in size once a sufficient number of games have been analyzed.

So the idea is to allow users (and scripts) to periodically scrape pgn's from fishtest, without having to keep the pgn's until the final conversion to json.

For final automation would need some tweaks to the download script, as this currently relies on downloaded pgn file names to avoid duplicates. Not sure if the script can be tweaked to only download all test that completed on the previous day, for example.

### DIFF
--- a/jsonmerge.py
+++ b/jsonmerge.py
@@ -1,0 +1,30 @@
+import argparse, json
+
+
+def merge_file(filename, merged_data):
+    with open(filename) as file:
+        data = json.load(file)
+        for key, value in data.items():
+            merged_data[key] = merged_data.get(key, 0) + value
+
+
+def main():
+    parser = argparse.ArgumentParser(description="Merge several JSON files.")
+    parser.add_argument("input", nargs="+", help="JSON files to merge")
+    parser.add_argument(
+        "-o", "--output", default="merged_data.json", help="merged JSON file"
+    )
+    args = parser.parse_args()
+
+    merged_data = {}
+    for filename in args.input:
+        merge_file(filename, merged_data)
+
+    with open(args.output, "w") as output_file:
+        json.dump(merged_data, output_file, indent=4)
+
+    print(f"Merging JSON file saved to {args.output}.")
+
+
+if __name__ == "__main__":
+    main()


vondele:
wouldn't this merging not be better to happen 'on the fly' in the analysis script, i.e. just give that analysis script multiple json files as an argument.

vondele:
concerning the downloads, yes, I think the download script should be improved to organize data a bit better, and ultimately, also the analysis script.

probably we should be organizing downloads as for example, a directory per day, and subdirectories for each test. 

Ultimately, we should fix a short-coming in the analysis script, and only analyse from master point of view (right now the assumption is that the opponent and master are sufficiently close so that they are captured with the same model. Holds most of the time at LTC I would say, except, for example, for regression tests). Ultimately, we would be interested in being able to track the master normalization evolution over time.

robertnurnberg:
Regarding your point on the analysis script: wouldn't the cleanest solution be to just sample the master evaluations when we convert pgn to json?

robertnurnberg:
Just for completeness: an alternative strategy to organize the pgn downloads in a cleaner dir structure would be the following workflow:
download all ltc tests that finished on the previous day -> convert these to json (and delete pgn's) -> merge json into master json (repeat daily in cron job)

robertnurnberg:
Thinking ahead to tracking the master normalization over time: the best solution would probably be to have one json file per `Base-sha` sf version? this would need be done at the pgn->json stage.

robertnurnberg:
Closing, as best way forward seems to be:

- in pgn -> json stage: filter for evaluations only from master (colour with `Base-sha` tag)
- in pgn -> json stage: by default or per cli option create one json file per `Base-sha` version
- in fitting script: allow several json files as input
- in addition organize directory structure for download script better

vondele:
I agree with the final comments, at least that seems the way to go. One thing though, if we can have a way to keep it simple for non-SF engines to use the code (as is now, just provide a bunch of pgns and use all evals) that would be good, it is nice that a few engines have started to adopt this normalization.

robertnurnberg:
Agreed. So new functionality best done with using non-default command line parameters. E.g. `--baseOnly` or similar.

PR #25, opened 2023-08-30
allow several json files as input

robertnurnberg:
No change with default parameters.

### DIFF
--- a/scoreWDL.py
+++ b/scoreWDL.py
@@ -10,9 +10,9 @@ parser = argparse.ArgumentParser(
 )
 parser.add_argument(
     "filename",
-    nargs="?",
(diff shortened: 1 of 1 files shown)


scoreWDL.py
```diff
@@ -87,9 +87,13 @@
 )
 fig.suptitle(title, fontsize="x-large")
 
-print(f"Reading score stats from {args.filename}.")
-with open(args.filename) as infile:
-    inputdata = json.load(infile)
+for filename in args.filename:
+    print(f"Reading score stats from {filename}.")
+    inputdata = {}
```
vondele, line 92: is that the correct place to zero stuff?

PR #26, opened 2023-08-31
allow regex filter for engine name

robertnurnberg:
Allow filtering of pgn data based on engine name. Sample usage: `./scoreWDLstat -o master.json --matchEngine "Base-.*"` to filter just master games and `./scoreWDLstat -o commit-12345.json --matchEngine ".*-12345"` to filter only games for a certain commit.

With some bash script magic, this would allow tracking of SF's WDL model in time.

PS: I also fixed the `Progress` indicator, so that now it starts at 1 and ends with `files_chunked.size()`.

### DIFF
--- a/scoreWDLstat.cpp
+++ b/scoreWDLstat.cpp
@@ -5,6 +5,7 @@
 #include <fstream>
 #include <iostream>
 #include <mutex>
+#include <regex>
 #include <string>
 #include <unordered_map>
 #include <vector>
@@ -53,7 +54,7 @@ struct std::hash<Key> {
 // unordered map to count (outcome, move, material, score) tuples in pgns
 using map_t = std::unordered_map<Key, int>;
 
-std::atomic<std::size_t> total_chunks = 0;
+std::atomic<std::size_t> total_chunks = 1;
 
 namespace analysis {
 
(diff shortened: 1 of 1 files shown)


scoreWDLstat.cpp
```diff
@@ -104,14 +105,34 @@ float fast_stof(const char *str) {
...
 /// @param game
-void ana_game(map_t &pos_map, const std::optional<Game> &game) {
+/// @param regexStr
+void ana_game(map_t &pos_map, const std::optional<Game> &game, const std::string &regexStr) {
     if (game.value().headers().find("Result") == game.value().headers().end()) {
         return;
     }
 
+    Color filterSide = Color::NONE;
+    if (regexStr != "") {
```
Disservin, line 118: c++ way would be to say
```cpp
if (!regexStr.empty())
```

scoreWDLstat.cpp
```diff
@@ -329,20 +353,21 @@ bool find_argument(const std::vector<std::string> &args,
...
 /// @param argc
-/// @param argv Possible ones are --dir, --file and -o
+/// @param argv Possible ones are --dir, --file, --matchEngine and -o
 /// @return
 int main(int argc, char const *argv[]) {
     const std::vector<std::string> args(argv + 1, argv + argc);
 
     std::vector<std::string> files_pgn;
-    std::string json_filename = "scoreWDLstat.json";
+    std::string regexStr = "", json_filename = "scoreWDLstat.json";
```
Disservin, line 370: You dont need to assign `""` to a string.

Disservin (commented):
current code style is that all variables are in snake case.

robertnurnberg:
Fixed style issues.

Disservin (approved):
lgtm

PR #27, opened 2023-09-03
allow recursive search for pgn files

robertnurnberg:
Will be needed once we go for new (sub)directory structure in fishtest pgn download directory.

No change with default parameters.


Disservin:
doesnt compile for me
```
g++ -Wall -Wextra -std=c++17 -O3 -march=native -o scoreWDLstat scoreWDLstat.cpp
scoreWDLstat.cpp: In function 'std::vector<std::__cxx11::basic_string<char> > get_files(const std::string&, bool)':
scoreWDLstat.cpp:253:53: error: invalid initialization of reference of type 'const std::string&' {aka 'const std::__cxx11::basic_string<char>&'} from expression of type 'const std::filesystem::__cxx11::path'
  253 |             auto subdir_files = get_files(entry.path(), true);
      |                                           ~~~~~~~~~~^~
scoreWDLstat.cpp:244:69: note: in passing argument 1 of 'std::vector<std::__cxx11::basic_string<char> > get_files(const std::string&, bool)'
  244 | [[nodiscard]] std::vector<std::string> get_files(const std::string &path, bool recursive = false) {
      |                                                  ~~~~~~~~~~~~~~~~~~~^~~~
make: *** [Makefile:22: all] Error 1
```

scoreWDLstat.cpp
```diff
@@ -239,13 +239,19 @@ void ana_files(map_t &map, const std::vector<std::string> &files, const std::str
...
 
     for (const auto &entry : fs::directory_iterator(path)) {
-        if (entry.path().extension() == ".pgn") {
-            files.push_back(entry.path().string());
+        if (fs::is_regular_file(entry)) {
+            if (entry.path().extension() == ".pgn") {
+                files.push_back(entry.path().string());
+            }
+        } else if (recursive && fs::is_directory(entry)) {
+            auto subdir_files = get_files(entry.path(), true);
```
Disservin, line 253: this needs to be `entry.path().string()` to compile (at least for me)

robertnurnberg:
Thanks for the quick review. Now fixed.

PR #28, opened 2023-09-03
allow date-test subdirs for fishtest pgns

robertnurnberg:
This PR creates the discussed path/date/testId/ directory structure. In addition:

- add `--page` cli option to harvest pgn's from older LTC test pages
- create `--path` directory if it does not exist yet
- use `urllib.request.urlretrieve` to simplify pgn download

### DIFF
--- a/download_fishtest_pgns.py
+++ b/download_fishtest_pgns.py
@@ -8,43 +8,70 @@ parser.add_argument(
     default="./",
     help="path to directory in which to store downloaded pgns",
 )
+parser.add_argument(
+    "--subdirs",
+    action="store_true",
+    help="use PATH/date/test-id/ (sub)directory structure",
+)
+parser.add_argument(
+    "--page",
+    type=int,
+    default=1,
+    help="page of LTC tests to download from",
+)
 args = parser.parse_args()
 if args.path == "":
     args.path = "./"
 elif args.path[-1] != "/":
     args.path += "/"
 
-print(f"Download pgns to directory {args.path} ...")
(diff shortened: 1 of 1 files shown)


download_fishtest_pgns.py
```diff
@@ -58,16 +85,6 @@
         time.sleep(0.1)  # be server friendly... wait a bit between requests
         url = "http://tests.stockfishchess.org/api/pgn/" + filename
         try:
-            response = urllib.request.urlopen(url)
+            urllib.request.urlretrieve(url, path + filename)
```
vondele, line 88: how does this deal with the encoding ? The 'weird stuff' that it replaces was to be able to deal with all fishtest pgns (in particular with some strange timezone characters, IIRC).

robertnurnberg: I wanted to avoid the encoding problems. My (limited) understanding is that this works like `wget`, i.e. a binary download of the file.

robertnurnberg: So my understanding is that master tries to somehow parse the .pgn files as text input, while this PR just downloads the file as is.

Disservin:
This reverts the changes we just merged in the C++ part?

robertnurnberg:
Good catch, sorry. Must have messed up some local rebasing. Let me fix this.

download_fishtest_pgns.py
vondele: and I guess downloading as is, is a reasonable thing. There might be some fall-out though. We'll see.

Issue #32, opened 2023-09-03
downloads fail without error message


robertnurnberg:
On a fresh clone, I get as output for `python download_fishtest_pgns.py --path pgns --subdirs --page 2`

```
Found 0 fully downloaded tests in pgns/ already.
Downloading pgns to pgns/23-08-30/64efc1bdb0db1f4c8581ee29/ ...
  Fetching 453 missing pgn files ...
Downloading pgns to pgns/23-08-31/64f0565fb0db1f4c8581ffc5/ ...
```

Then killing the process and checking the directory `pgns/23-08-30/64efc1bdb0db1f4c8581ee29` gives an empty directory, whereas the second directory starts to get filled.

Will try to investigate tomorrow.

robertnurnberg:
Ok, so the reason is that the pgn no longer exists: https://tests.stockfishchess.org/api/pgn/64efc1bdb0db1f4c8581ee29-0.pgn.

While http://tests.stockfishchess.org/api/pgn/64f0565fb0db1f4c8581ffc5-0.pgn works.

Will add some error output in the script and then close the issue.

PS: For how long are pgn's stored on fishtest?

vondele:
first, I believe not every task will have a pgn stored, so it will somehow need to deal with that. 

The pgns are not stored for a very long time, LTC runs are stored for longer. https://github.com/official-stockfish/fishtest/blob/master/server/utils/purge_pgn.py

PR #33, opened 2023-09-04
improve error handling

robertnurnberg:
Output errors we encounter when trying to download pgn files. Skip remaining downloads for a test as soon as we get an HTTP error code.

Fixes https://github.com/official-stockfish/WDL_model/issues/32.

### DIFF
--- a/download_fishtest_pgns.py
+++ b/download_fishtest_pgns.py
@@ -1,7 +1,10 @@
 import urllib.request, urllib.error, urllib.parse
 import argparse, time, re, os
 
-parser = argparse.ArgumentParser()
+parser = argparse.ArgumentParser(
+    description="Download pgns from completed LTC tests on fishtest.",
+    formatter_class=argparse.ArgumentDefaultsHelpFormatter,
+)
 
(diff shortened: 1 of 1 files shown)


vondele:
as mentioned elsewhere, I think there are cases where a pgn might be missing for 1 task, but pgns for other tasks are present (like when a worker would have a problem sending the pgn, or is stopped before).

robertnurnberg:
So you prefer not to `break`?

robertnurnberg:
For the example from https://github.com/official-stockfish/WDL_model/issues/32 _all_ the pgn's are missing, I think.

robertnurnberg:
Could have a counter, say 3, and break once 3 pgn's are missing?

robertnurnberg:
With the latest change a third missing task per test means the test gets skipped. cli option `--leniency` can be used to be more or less lenient. 

Also made `--help` a bit more informative.

robertnurnberg:
Final version is more lenient: only count _consecutive_ HTTP errors, and allow 3 of those before skipping.

PR #38, opened 2023-09-10
add chess variant to subdir structure

robertnurnberg:
Not sure if this is the right way to go, hence draft for now.

This PR would allow to separate the pgn data based on chess variant played, without any additional changes to the other scripts in this repo.

Initially I thought FRC and DFRC do not use depth 8 books, and so this would have been necessary to get the move count information right. Having checked that also FRC and DFRC seem to use 8-move books, this PR is less urgent from that point of view.

The script has not been tested with actual FRC or DFRC LTC tests, as I could not easily find a page of the form https://tests.stockfishchess.org/tests/finished?ltc_only=1&page=1 that contains such tests.

### DIFF
--- a/download_fishtest_pgns.py
+++ b/download_fishtest_pgns.py
@@ -14,7 +14,7 @@ parser.add_argument(
 parser.add_argument(
     "--subdirs",
     action="store_true",
-    help="use PATH/date/test-id/ (sub)directory structure",
+    help="use PATH/variant/date/test-id/ (sub)directory structure",
 )
 parser.add_argument(
     "--page",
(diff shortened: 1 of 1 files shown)


robertnurnberg:
Just putting this here as a thought: one could also replace `variant` with `book` for even more fine-grained storage.

vondele:
I don't think this should be encoded in the directory naming. This probably should go in some metadata stored in a file, since we have already a directory per test. Such a file could easily grow in content (e.g. book, game manager, TC, game outcome, etc) which the analysis script could be using more flexibly.

robertnurnberg:
Agreed, will make a start based on the parsing code in this PR.

PR #40, opened 2023-09-10
store meta data in local json files

robertnurnberg:
Additional data to store can be added in future, or based on comments to this PR already now.

### DIFF
--- a/download_fishtest_pgns.py
+++ b/download_fishtest_pgns.py
@@ -84,6 +84,7 @@ for test, dateStr in ids:
         "new_net",
         "new_options",
         "new_tc",
+        "sprt",
         "tc",
         "threads",
     ]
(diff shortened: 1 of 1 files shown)


vondele:
looks good, possible further fields from the page could be `start time` `last updated` and if this is `sprt` or not.

robertnurnberg:
So `sprt` was easy, the other two need more time, as the data is encoded slightly differently in `webContent`. I suggest to merge this and add PR on top.

vondele:
actually, it might be good to store the base and new sha's as well as the branch used for the test in the metadata as well. While it is part of the header, it might be convenient to have it there.

One thing that could be done for example automatically with this is to extract the pawn normalization, by accessing the proper file in the git repo.

PR #42, opened 2023-09-16
cleanup scoreWDL.py

Disservin:
First cleanup patch of the python code. Group logic and remove usage of any "globals".
I am pretty open to changes, concering naming or the way I divided the functions.

This serves as a first step for later refactors.

### DIFF
--- a/scoreWDL.py
+++ b/scoreWDL.py
@@ -4,133 +4,80 @@ from ast import literal_eval
 from scipy.interpolate import griddata
 from scipy.optimize import curve_fit
 
-parser = argparse.ArgumentParser(
-    description="Fit Stockfish's WDL model to fishtest game statistics.",
-    formatter_class=argparse.ArgumentDefaultsHelpFormatter,
-)
-parser.add_argument(
-    "filename",
-    nargs="*",
-    help="json file(s) with fishtest games' WDL statistics",
-    default=["scoreWDLstat.json"],
-)
-parser.add_argument(
-    "--NormalizeToPawnValue",
-    type=int,
-    default=328,
-    help="Value needed for converting the games' cp scores to the SF's internal score.",
-)
-parser.add_argument(
-    "--moveMin",
-    type=int,
-    default=0,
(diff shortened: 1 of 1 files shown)


Disservin:
I just set the PR status to draft to wait for input and proper testing. So far `python scoreWDL.py` seems to run fine.

vondele:
had a first look, the code style looks good to me. I've done a first check that it returns the same text output as master with all default arguments.

scoreWDL.py
```diff
@@ -4,384 +4,491 @@
...
-            model_ms, poly3(model_ms, *popt_bs), "m-", label="fit: " + label_bs
+        if args.fit:
+            title = "Summary of win-draw-loss model analysis"
+            pgnName = "WDL_model_summary.png"
+        else:
+            title = "Summary of win-draw-loss data"
+            pgnName = f"WDL_data_{args.yData}.png"
+
+        plot = WdlPlot(
+            "Summary of win-draw-loss model analysis", "WDL_model_summary.png"
```
robertnurnberg, line 468: to be consistent with master, this line should use `title` and `pgnName`

scoreWDL.py
```diff
@@ -4,384 +4,491 @@
...
+        data_loader.load_json(),
+        args.moveMin,
+        args.moveMax,
+        args.NormalizeToPawnValue,
+        args.yData,
+    )
+
+    xs, ys, zwins, zdraws, zlosses = data_loader.convert_to_model(win, draw, loss)
+
+    wdl_model = WdlModel(args, plot)
```
robertnurnberg, line 485: plot here could be undefined

maybe a better logic would be to make `plot` a member data of `WdlModel`

robertnurnberg:
Looks very nice overall.

scoreWDL.py
Disservin: oh thanks, I didnt notice I hardcoded those :p

robertnurnberg:
lgtm

PR #43, opened 2023-09-18
second cleanup of scoreWDL.py

Disservin:
- plot and model logic is now a bit more isolated
- small refactor of `zwins[i] = fit.wdl(xs[i], ys[i], popt_as, popt_bs)[0] / 1000.0` this was actually overwriting the initial array, now creates a temporary one
- made normalized_axis static

### DIFF
--- a/scoreWDL.py
+++ b/scoreWDL.py
@@ -3,6 +3,7 @@ from collections import Counter
 from ast import literal_eval
 from scipy.interpolate import griddata
 from scipy.optimize import curve_fit
+from dataclasses import dataclass
 
 
 class WdlPlot:
@@ -80,13 +81,7 @@ class DataLoader:
             zwins.append(win[x, y] / total)
             zdraws.append(draw[x, y] / total)
             zlosses.append(loss[x, y] / total)
-        return {
-            "xs": xs,
-            "ys": ys,
-            "zwins": zwins,
-            "zdraws": zdraws,
-            "zlosses": zlosses,
-        }
+        return RawModelData(xs, ys, zwins, zdraws, zlosses)
 
 
 #
@@ -138,6 +133,26 @@ class ModelFit:
         l = int(1000 * ModelFit.winmodel(-score, a, b))
         d = 1000 - w - l
         return w, d, l
+
+
+@dataclass
+class RawModelData:
+    xs: list
+    ys: list
+    zwins: list
+    zdraws: list
+    zlosses: list
+
+
+@dataclass
+class ModelData:
+    popt_as: list
+    popt_bs: list
+    model_ms: list
+    model_as: list
+    model_bs: list
+    label_as: str
+    label_bs: str
 
 
 class WdlModel:
@@ -276,17 +291,11 @@ class WdlModel:
             % tuple(popt_bs)
         )
 
-        return {
-            "popt_as": popt_as,
-            "popt_bs": popt_bs,
-            "model_ms": model_ms,
-            "model_as": model_as,
-            "model_bs": model_bs,
-            "label_as": label_as,
-            "label_bs": label_bs,
-        }
+        return ModelData(
+            popt_as, popt_bs, model_ms, model_as, model_bs, label_as, label_bs
+        )
 
-    def create_plot(self, raw_model_data, model):
+    def create_plot(self, raw_model_data: RawModelData, model):
         # graphs of a and b as a function of move/material
         print("Plotting move/material dependence of model parameters.")
 
@@ -334,12 +343,10 @@ class WdlModel:
         xmax = ((3 * self.args.NormalizeToPawnValue) // 100 + 1) * 100
         ymin, ymax = self.args.yPlotMin, self.args.yDataMax
         grid_x, grid_y = np.mgrid[xmin:xmax:30j, ymin:ymax:22j]
-        points = np.array(list(zip(raw_model_data["xs"], raw_model_data["ys"])))
+        points = np.array(list(zip(raw_model_data.xs, raw_model_data.xs)))
 
         # data
-        zz = griddata(
-            points, raw_model_data["zwins"], (grid_x, grid_y), method="linear"
-        )
+        zz = griddata(points, raw_model_data.zwins, (grid_x, grid_y), method="linear")
         cp = self.plot.axs[0, 1].contourf(grid_x, grid_y, zz, contourlines)
         self.plot.fig.colorbar(cp, ax=self.plot.axs[:, -1], shrink=0.618)
         CS = self.plot.axs[0, 1].contour(
@@ -353,13 +360,13 @@ class WdlModel:
         # model
         if self.args.fit:
             zwins = []
-            for i in range(0, len(raw_model_data["xs"])):
+            for i in range(0, len(raw_model_data.xs)):
                 zwins.append(
                     fit.wdl(
-                        raw_model_data["xs"][i],
-                        raw_model_data["ys"][i],
-                        model["popt_as"],
-                        model["popt_bs"],
+                        raw_model_data.xs[i],
+                        raw_model_data.ys[i],
+                        model.popt_as,
+                        model.popt_bs,
                     )[0]
                     / 1000.0
                 )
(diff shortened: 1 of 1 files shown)


Disservin:
for some parts I was a bit unsure about the variable names, feel free to suggest some different ones...

scoreWDL.py
```diff
@@ -76,7 +80,13 @@ def convert_to_model(self, win, draw, loss):
             zwins.append(win[x, y] / total)
             zdraws.append(draw[x, y] / total)
             zlosses.append(loss[x, y] / total)
-        return xs, ys, zwins, zdraws, zlosses
+        return {
```
robertnurnberg, line 83: I am not a big fan of these returned dictionaries.

Here a more natural approach might be to make `raw_model_data` a class, whose constructor takes the three arguments, and then modifies member data `xs` etc.

Disservin: Mh kinda decided against classes because they wont have any behaviour/methods and just serve as a data wrapper in which case they they simply add more complexity imo

vondele:
for me this looks good, needs rebasing and undrafting. I have no strong preference for the return style, if some concensus could be reached there I could merge this.

robertnurnberg:
My concern with the proposed return style is that it adds another layer of abstraction. And a typo anywhere in the keyword of the returned dict will only throw a (dict keyword) error once the code is executed in that branch. Whereas dealing with the variables directly (e.g. as member data in a class) would lead to an immediate interpreter error in case of a typo.

Having said all that, I do appreciate @Disservin's work on cleaning up the code. And so if he is happy with the PR as it stands, I am also fine with merging this.

Disservin:
@robertnurnberg I am now making use of python's `dataclass`decorator to store the return type, this provides some type safety and is more concise than also adding the init function. 
https://docs.python.org/3/library/dataclasses.html

This raises the required python version to 3.7, I dont think that's a problem?

Disservin:
Also as the third refactoring, I would any way like to type (hint) the entire code anyway.

robertnurnberg:
> @robertnurnberg I am now making use of python's `dataclass`decorator to store the return type, this provides some type safety and is more concise than also adding the init function. https://docs.python.org/3/library/dataclasses.html

Nice!

> This raises the required python version to 3.7, I dont think that's a problem?

I think even vondele's rasperry pi's have python 3.7.3. :) So no problem as far as I can see.

robertnurnberg:
Did someone check this for non-functionality? I didn't and now found a problem.

For the attached `updateWDL.json` I get an error for `python scoreWDL.py updateWDL.json` in master, but not in the revision before this commit.

If someone can confirm this, can we revert? (It holds up my open PR. ;)  )

[updateWDL.json.gz](https://github.com/official-stockfish/WDL_model/files/12777768/updateWDL.json.gz)

Disservin:
What error do you get ? I haven’t but I think vondele did or at least thats what I understood. Maybe it’s also simply faster to fix this rather than revert and fix?

robertnurnberg:
The fitting breaks down with an error.

To see if you get the same, you need to unzip the JSON file I uploaded and run the given commands for the two commits to see if you get different behaviour.

Disservin:
I mean it would be helpful if you just show the error that you get ;) as I currently can’t test it for a couple of hours

robertnurnberg:
> I mean it would be helpful if you just show the error that you get ;) as I currently can’t test it for a couple of hours

```
Converting scores with NormalizeToPawnValue = 328.
Reading score stats from updateWDL.json.
Retained (W,D,L) = (1509055, 3676666, 1525452) positions.
Fit WDL model based on move.
const int NormalizeToPawnValue = 336;
Corresponding spread = 59;
Corresponding normalized spread = 0.17549117532825584;
Draw rate at 0.0 eval at move 32 = 0.9933190235366156;
as = ((-1.152 * x / 32 + 8.011) * x / 32 + -10.002) * x / 32 + 339.961
bs = ((-6.121 * x / 32 + 39.089) * x / 32 + -71.393) * x / 32 + 97.532
     constexpr double as[] = {  -1.15191932,    8.01059533,  -10.00197382,  339.96079703};
     constexpr double bs[] = {  -6.12070895,   39.08942305,  -71.39252269,   97.53230739 };
Plotting move/material dependence of model parameters.
Processing done, plotting 2D data.
  File "/tmp/rn/WDL_model/scoreWDL.py", line 553, in <module>
    wdl_model.create_plot(
  File "/tmp/rn/WDL_model/scoreWDL.py", line 349, in create_plot
    zz = griddata(points, raw_model_data.zwins, (grid_x, grid_y), method="linear")
  File "/usr/lib/python3/dist-packages/scipy/interpolate/_ndgriddata.py", line 260, in griddata
    ip = LinearNDInterpolator(points, values, fill_value=fill_value,
  File "interpnd.pyx", line 280, in scipy.interpolate.interpnd.LinearNDInterpolator.__init__
  File "_qhull.pyx", line 1841, in scipy.spatial._qhull.Delaunay.__init__
  File "_qhull.pyx", line 357, in scipy.spatial._qhull._Qhull.__init__
scipy.spatial._qhull.QhullError: QH6154 Qhull precision error: Initial simplex is flat (facet 1 is coplanar with the interior point)

While executing:  | qhull d Qc Qt Qbb Q12 Qz
Options selected for Qhull 2019.1.r 2019/06/21:
  run-id 2091095942  delaunay  Qcoplanar-keep  Qtriangulate  Qbbound-last
  Q12-allow-wide  Qz-infinity-point  _pre-merge  _zero-centrum  Qinterior-keep
  Pgood  _max-width 2.6e+03  Error-roundoff 1.8e-12  _one-merge 1.3e-11
  Visible-distance 3.6e-12  U-max-coplanar 3.6e-12  Width-outside 7.3e-12
  _wide-facet 2.2e-11  _maxoutside 1.5e-11

The input to qhull appears to be less than 3 dimensional, or a
computation has overflowed.

Qhull could not construct a clearly convex simplex from points:
- p17438(v4): -0.21 -0.21 1.3e+03
- p8667(v3):     0     0     0
- p17338(v2): 1.3e+03 1.3e+03 1.2e+03
- p0(v1): -1.3e+03 -1.3e+03 1.2e+03

The center point is coplanar with a facet, or a vertex is coplanar
with a neighboring facet.  The maximum round off error for
computing distances is 1.8e-12.  The center point, facets and distances
to the center point are as follows:

center point -0.05361 -0.05361    924.4

facet p8667 p17338 p0 distance=    0
facet p17438 p17338 p0 distance=    0
facet p17438 p8667 p0 distance=    0
facet p17438 p8667 p17338 distance=    0

These points either have a maximum or minimum x-coordinate, or
they maximize the determinant for k coordinates.  Trial points
are first selected from points that maximize a coordinate.

The min and max coordinates for each dimension are:
  0:     -1312      1312  difference= 2624
  1:     -1312      1312  difference= 2624
  2:         0      1312  difference= 1312

If the input should be full dimensional, you have several options that
may determine an initial simplex:
  - use 'QJ'  to joggle the input and make it full dimensional
  - use 'QbB' to scale the points to the unit cube
  - use 'QR0' to randomly rotate the input for different maximum points
  - use 'Qs'  to search all points for the initial simplex
  - use 'En'  to specify a maximum roundoff error less than 1.8e-12.
  - trace execution with 'T3' to see the determinant for each point.

If the input is lower dimensional:
  - use 'QJ' to joggle the input and make it full dimensional
  - use 'Qbk:0Bk:0' to delete coordinate k from the input.  You should
    pick the coordinate with the least range.  The hull will have the
    correct topology.
  - determine the flat containing the points, rotate the points
    into a coordinate plane, and delete the other coordinates.
  - add one or more points to make the input full dimensional.
```

So the error seems not in the fitting, but in the plotting.

robertnurnberg:
https://github.com/official-stockfish/WDL_model/pull/56 fixes this for me, all good. :)

robertnurnberg:
Just as a heads up for the future: whenever I do refactorings, I keep local test input data that I run virtually _any_ change I make to the source for, to see if the output is unchanged. (For the .py script in this repo that would be a .json input file and the .png output, which can be checked with `diff` against the master output.) Then before I open the PR I also test corner cases, and all the possible cli parameters. I think this is good practice.

PR #45, opened 2023-09-24
enable automatic WDL change detection

robertnurnberg:
The new script `updateWDL.sh` allows automation for the detection of changes in the WDL model.

In its current setting, it checks only data since the commit with the recent NNUE arch change.

Sooner or later we should update the WDL model in SF (to allow updates for the cdb workers, for example), and this PR makes this easier.

In future, the script could be changed to fully automated checking WDL changes and opening PR's on SF for the required changes etc.

### DIFF
--- a/updateWDL.sh
+++ b/updateWDL.sh
@@ -1,0 +1,74 @@
+#!/bin/bash
+
+# exit on errors
+set -e
+
+echo "started at: " `date`
+
+# give range of commit SHAs to be considered for the WDL fitting
+firstrev=70ba9de85cddc5460b1ec53e0a99bee271e26ece
+lastrev=HEAD
+
+# regex for book name
+bookname=UHO_4060_v3.epd
+
+# path for PGN files
+pgnpath=pgns
+
+# clone SF if needed
+if [[ ! -e Stockfish ]]; then
+    git clone https://github.com/official-stockfish/Stockfish.git >& clone.log
+fi
+
+# compile scoreWDLstat if needed
+make >& make.log
+
+# update SF and get a revision list
+cd Stockfish
+git checkout master >& checkout.log
+git fetch origin >& fetch.log
+git pull >& pull.log
+revs=`git rev-list $firstrev^..$lastrev`
+
+# get the currently valid value of NormalizeToPawnValue
+oldpawn=`git grep 'const int NormalizeToPawnValue' $firstrev -- src/uci.h | grep -oP 'const int NormalizeToPawnValue = \K\d+'`
+oldepoch=`git show --quiet --format=%ci $firstrev`
+newepoch=`git show --quiet --format=%ci $lastrev`
+
+# build a regex pattern to match all revisions
+regex_pattern=""
+for rev in $revs; do
+    regex_pattern="${regex_pattern}.*-$rev|"
+    newpawn=`git grep 'const int NormalizeToPawnValue' $rev -- src/uci.h | grep -oP 'const int NormalizeToPawnValue = \K\d+'`
+    if [[ $oldpawn -ne $newpawn ]]; then
+        echo "Revision $rev has wrong NormalizeToPawnValue ($newpawn != $oldpawn)"
+        exit 1
+    fi
+done
+
+# remove the trailing "|"
+regex_pattern="${regex_pattern%|}"
+
+cd ..
+
+echo "Look recursively in directory $pgnpath for games from SPRT tests using"\
+     "books matching \"$bookname\" for SF revisions between $firstrev (from"\
+     "$oldepoch) and $lastrev (from $newepoch)."
+
+# obtain the WDL data from games of SPRT tests of the SF revisions of interest
+./scoreWDLstat --dir $pgnpath -r --matchEngine $regex_pattern --matchBook "$bookname" --fixFEN --SPRTonly -o updateWDL.json >& scoreWDLstat.log
+
+# fit the new WDL model, keeping anchor at move 32
+python scoreWDL.py updateWDL.json --plot save --yDataTarget 32 --yDataMin 11 --yDataMax 120 --yPlotMin 18 --NormalizeToPawnValue $oldpawn >& scoreWDL.log
+
+# extract the total number of positions, and the new NormalizeToPawnValue
+poscount=`awk -F '[() ,]' '/Retained \(W,D,L\)/ {sum = 0; for (i = 9; i <= NF; i++) sum += $i; print sum; exit}' scoreWDL.log`
+newpawn=`grep -oP 'const int NormalizeToPawnValue = \K\d+' scoreWDL.log`
+
+if [[ $newpawn -ne $oldpawn ]]; then
+    echo "Based on $poscount positions, NormalizeToPawnValue should change from $oldpawn to $newpawn."
+else
+    echo "Based on $poscount positions, NormalizeToPawnValue should stay at $oldpawn."
+fi
+
+echo "ended at: " `date`
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,9 @@
 *.png
 *.json
 *.pgn
+*.log
 
 pgns*
 scoreWDLstat
 scoreWDLstat.exe
+Stockfish


updateWDL.sh
```diff
@@ -0,0 +1,63 @@
...
+# build a regex pattern to match all revisions
+regex_pattern=""
+for rev in $revs; do
+    regex_pattern="${regex_pattern}.*-$rev|"
+done
+
+# remove the trailing "|"
+regex_pattern="${regex_pattern%|}"
+
+# test if any PGN file contains the move counter "0 1"
```
vondele, line 42: that would only be right if the book has that property, so that's a property of the particular test.

vondele: (could be stored in the .json metadata file).

updateWDL.sh
```diff
@@ -0,0 +1,63 @@
...
+if [[ ! -e scoreWDLstat ]]; then
+    make >& make.log
+fi
+
+# update SF and get a revision list
+cd Stockfish
+git checkout master >& checkout.log
+git fetch origin >& fetch.log
+git pull >& pull.log
+revs=`git rev-list $firstrev^..$lastrev`
```
vondele, line 30: Here it would be really helpful to get and check NormalizeToPawnValue from these revisions, check if they are all the same, and use it as an argument to scoreWDL.py

robertnurnberg:
Thanks for the quick feedback. This is obviously a first version that can be built upon in future. For now it does what is needed for the next wdl model fit in SF, I think.

Regarding the move counter warning, it is a warning only and up to the user to fix the situation if and how they want to. Overall this is a transient issue, which will disappear as soon as fishtest employs a new game manager. But if we wanted to, we could do the pgn edits within this script, based on book info in JSON data. As this is a significant change, I would prefer to leave it for a future PR 

About getting `NormalizeToPawnValue` from the SF commits, I thought about it. But (a) would this come at an additional computational cost, and (b) the biggest need I see for now is to update the wdl model in SF using recent game data. And there the constant will be fixed.

What we could do is get the constant from the SF code and check for consistency in all the analysed commits. Again as a first step, with generalisations possible later on.

vondele:
I would not change downloaded pgns, keep them as 'original data'.

One could do something like 
`git grep 'NormalizeToPawnValue' 70ba9de85cddc5460b1ec53e0a99bee271e26ece -- uci.h`
on all sha's in the range, to check consistency.

robertnurnberg:
Agreed about `NormalizeToPawnValue`, I can look at that tonight.

Anything else regarding `NormalizeToPawnValue` is not practical at the moment. E.g. allowing variable `NormalizeToPawnValue` in the PGN data would require major changes in the other scripts.

Regarding the downloaded PGNs, the suggested fix in the warning is what I do in my cronjob at the moment. As far as I know, all the books used for LTC on fishtest are 8-move at the moment? It would be a pity to invest a lot of work (dict of books in use on fishtest with names and book length, getting book name from json data, fixing the .cpp code to allow book-exit offset), if the issue will go away on its own eventually. (It only arises because fishtest at present does not output correct .pgn data.)

vondele:
> Agreed about `NormalizeToPawnValue`, I can look at that tonight.

Thanks

> Anything else regarding `NormalizeToPawnValue` is not practical at the moment. E.g. allowing variable `NormalizeToPawnValue` in the PGN data would require major changes in the other scripts.

I agree, no need to allow for mixed values right now.

> 
> Regarding the downloaded PGNs, the suggested fix in the warning is what I do in my cronjob at the moment. As far as I know, all the books used for LTC on fishtest are 8-move at the moment? It would be a pity to invest a lot of work (dict of books in use on fishtest with names and book length, getting book name from json data, fixing the .cpp code to allow book-exit offset), if the issue will go away on its own eventually. (It only arises because fishtest at present does not output correct .pgn data.)

DFRC book starts at 'startpos' for example. If we run with other books (e.g. noob_3moves) it is different, or if we use pgn books instead of epd books it is sometimes correct (i.e. those that store the move sequence in the pgn). So, while the bulk of the pgns could be fixed, others would be turned from correct to wrong.

robertnurnberg:
> DFRC book starts at 'startpos' for example. If we run with other books (e.g. noob_3moves) it is different, or if we use pgn books instead of epd books it is sometimes correct (i.e. those that store the move sequence in the pgn). So, while the bulk of the pgns could be fixed, others would be turned from correct to wrong.

Right, that is a problem. The fix to this would be in the .cpp script, agreed? That would now have to check for the JSON data and then adjust the full move counter depending on the advertised book.

vondele:
> > DFRC book starts at 'startpos' for example. If we run with other books (e.g. noob_3moves) it is different, or if we use pgn books instead of epd books it is sometimes correct (i.e. those that store the move sequence in the pgn). So, while the bulk of the pgns could be fixed, others would be turned from correct to wrong.
> 
> Right, that is a problem. The fix to this would be in the .cpp script, agreed? That would now have to check for the JSON data and then adjust the full move counter depending on the advertised book.

I think so.

updateWDL.sh
```diff
@@ -0,0 +1,74 @@
...
+# path for PGN files
+pgnpath=pgns
+
+# clone SF if needed
+if [[ ! -e Stockfish ]]; then
+    git clone https://github.com/official-stockfish/Stockfish.git >& clone.log
+fi
+
+# compile scoreWDLstat if needed
+if [[ ! -e scoreWDLstat ]]; then
```
vondele, line 21: probably safer to always call make, and let make decide if it needs to rebuild?

robertnurnberg: D'oh! Of course. :)

(This is as good a time as any to say that I really appreciate your kind and professional way to communicate, on display since your very first debates with Marco when you joined the project.)

updateWDL.sh
```diff
@@ -0,0 +1,72 @@
...
+    echo "WARNING: Found at least one PGN file with move counter 0 1."
+    echo "(In the absence of DFRC data this indicates wrong move counters.)"
+    echo "(Issue should be fixed on fishtest (long term) or in scoreWDLstat.cpp (short term)."
+fi
+
+# obtain the filtered WDL data
+./scoreWDLstat --dir $pgnpath -r --matchEngine $regex_pattern -o updateWDL.json >& scoreWDLstat.log
+
+# fit the new WDL model, using 8-move offsets in parameters for consistency
+python scoreWDL.py updateWDL.json --plot no --yDataTarget 40 --yDataMin 11 --yDataMax 128 --NormalizeToPawnValue $oldpawn >& scoreWDL.log
```
robertnurnberg, line 60: So here I would need some input. @vondele 

1) Do we want to exclude (D)FRC data? This is now possible, and I think it may be a good idea. So that the WDL model is fitted just for classical chess, and (D)FRC users would need to keep that in mind.

2) I have shifted _all_ the parameters by 8 moves. But I believe you would prefer to keep the anchor at 32, which makes sense to avoid user confusion and keep documentation and SF code itself unchanged. So I suggest to move the anchor to 32, leaving the other shifted values in place. OK?

vondele: 1. I would just restrict to the UHO_4060_v3.epd book for now, that covers 99% of cases.
2. Yes, let's keep the anchor at 32 moves. Also the other parameters are not sacred, if it makes sense to make adjustments we can do that.

Independent from these question, but while we are at it, it would make sense to exclude also progression tests from contributing to the data, i.e. use only the SPRT tests and drop fixed number of games tests, at least until we figure out how to take the Elo difference between players into account.

robertnurnberg:
>Yes, let's keep the anchor at 32 moves. Also the other parameters are not sacred, if it makes sense to make adjustments we can do that.

Agreed. I have kept the offset for the lower limits, as there must have been good reasons for you to set them at 3 and 10, respectively. But before we open the next `update winrate model` PR on SF, we could do some checks to see if they can be reduced without harm. To keep in mind: as the lower fit-limit is 11 moves at present, it would make sense also in SF to clamp the move number to [11, 120] before evaluating the WDL model.

>Independent from these question, but while we are at it, it would make sense to exclude also progression tests from contributing to the data, i.e. use only the SPRT tests and drop fixed number of games tests, at least until we figure out how to take the Elo difference between players into account.

Agreed. Is the easiest way to determine this if the `"sprt"` key is in the json data? How best to design the cli for the .cpp code for that? Just `--sprtonly`?

vondele:
yes, I think we should have a careful look prior to merging the updated model based on this way of analysis if something is needed, e.g. clamping. Since we have now most data around 1.0 at move 8, the model risks to be not too accurate at move 8. Eventually we'll figure out something :-)

yeah, I think sprt in the json is the trick, you can check that a test like https://tests.stockfishchess.org/tests/view/650dcf1bfb151d43ae6d704f is indeed excluded.

robertnurnberg:
The only way to fix that "hole" before move 8 is to have some bookless data, I guess.

vondele:
yes, I think it would need runs with very short books, and more varied openings, that's not something we have or do right now. Anyway, we'll see. The other solution would be e.g. to use material based WDL models since even the 8 move books cover all material values.

vondele:
BTW, undraft the PR once you think it is ready for merging..

robertnurnberg:
This PR still needs work on the .cpp script to deal with the move counter (which I plan to do over the next few days). Will undraft once all that is done. :)

robertnurnberg:
> yes, I think we should have a careful look prior to merging the updated model based on this way of analysis if something is needed, e.g. clamping. Since we have now most data around 1.0 at move 8, the model risks to be not too accurate at move 8. Eventually we'll figure out something :-)

As we are already looking ahead to the next `update winrate model` PR for SF, I wanted to just put this here for reference. I believe the correct way for [this](https://github.com/official-stockfish/Stockfish/blob/22cdb6c1ea1f5ca429333bcbe26706c8b4dd38d7/src/uci.cpp#L212C5-L212C5) line in `uci.cpp` is
```cpp
double m = std::clamp(ply / 2 + 1, 11, 120) / 32.0;
```
(This is for the new clamping window. For master it should be clamped to `3, 120`, but best to fix this together with the full move counter shift in the next update.)

robertnurnberg:
Undrafted after the merge of the recent PRs. Ready for review.

vondele:
LGTM

vondele:
ran this locally, and it worked very well. This is good progress, thanks!

robertnurnberg:
So locally I get 
```
started at:  Mon 2 Oct 07:46:23 CEST 2023
Look recursively in directory pgns for games from SPRT tests using books matching "UHO_4060_v3.epd" for SF revisions between 70ba9de85cddc5460b1ec53e0a99bee271e26ece (from 2023-09-22 19:26:16 +0200) and HEAD (from 2023-09-29 22:30:27 +0200).
Based on 220911561 positions, NormalizeToPawnValue should change from 328 to 339.
ended at:  Mon 2 Oct 07:47:59 CEST 2023
```

with the attached image. Out of curiosity, how many positions do we need to have a model robust enough to be considered for merging into SF?

WDL_model_summary

vondele:
I have never quantified how much we really need, mostly looking at the contour plots and going 'by eye'. The above plot is probably good enough, and it is about 150M positions. Last update to SF was 748M, the one before not documented, the one before 400M. Usually, I've been collecting data for about a month, but hat might be too much, in the end we're only fitting relatively few parameters. Just the normalization will need less data than the full model as well.

The new setup should allow for exploring that a bit more systematically, e.g. analyzing only every Nth game, and see what the variance is of the normalization constant between N independent samples created in this way. 

I think the larger error is now due to model used, i.e. the polynomial approximation is not too accurate for the as and bs (probably a rational function would do better), even though both as and bs have become increasingly constant over time. Now, it is unclear if it is worthwhile do that more accurately, the model has its limitations nevertheless (e.g. even for self-play, the TC dependence of especially 'b' is not taken into account). Also, some of the structure in the data (e.g. the asymmetry near move 20 or below), is most likely due to the fact that the opening book is UHO, and so samples with negative eval there are rare, and 'special'. 

I have been avoiding to update the model too frequently, if the normalization constant is not too much off (e.g. less than 10%), there is no pressure to do so.

robertnurnberg:
Ok, understood. As outlined above, one of my motivations was to allow the cdb worker to be updated. Especially the larger nets may give more accurate evaluations of drawn positions near the cdb leafs etc. But I agree that SF should do what is best for SF, and there less frequent changes to the WDL model is probably for the best.

Regarding the number of positions: my example has about 220M (not 150M?), agreed? I guess this would have been more than 440M in old currency, since now we ignore the non-master evals. In addition, the old data may have contained PTs, which we now ignore.

One final thought: the current WDL model of SF has this 8-move mismatch. So this may be a reason to update the model in the near future, even if the normalization constant changed by less than 10%.

robertnurnberg:
> Also, some of the structure in the data (e.g. the asymmetry near move 20 or below), is most likely due to the fact that the opening book is UHO, and so samples with negative eval there are rare, and 'special'.

Is not having negative (from white's PoV) evals in the data really a problem? I would need to check the details, but is there not (some) symmetry in the WDL (meaning that e.g. eval of Xcp gives same W as -Xcp gives for L?)

vondele:
150M referred to what I had locally, probably a little less than what you already downloaded, and indeed that's probably better than what we had. Fixing the issue with the 8 moves mismatch (as well as renormalize after the net arch change) would be good reasons. In the situation we have now, I would usually wait a little bit if somebody came with some rescaling of eval after the net change, and if not start collecting data for a couple of weeks. So its a good time to see if anything in the fitting process needs fine-tuning or if we're good to go.

vondele:
The model assumes symmetry indeed, now the data shows there is asymmetry (see top right graph near move 20), and trying to fit that is of course not possible.

Disservin:
Now I haven’t been following much of your talk recently but it’s nice to see good progress ;) I think the readme might need another update ?

robertnurnberg:
> 150M referred to what I had locally, probably a little less than what you already downloaded, and indeed that's probably better than what we had.

These are my local .crontab entries:
```
10 3 * * * cd /home/rn/git/wdl_model && python3 download_fishtest_pgns.py --leniency 10
10 5 * * * cd /home/rn/git/wdl_model && python3 download_fishtest_pgns.py --leniency 10 --page 2
 ```
 
 I found the default leniency a bit too strict, and for the cronjob I do not care about wasting time trying to download non-existent pgns.

robertnurnberg:
> The model assumes symmetry indeed, now the data shows there is asymmetry (see top right graph near move 20), and trying to fit that is of course not possible.

True, good spot. I think this may be one side-effect of the very sparse data at low move counts: no data for move count < 8, and virtually no data around move 8 with evals in (clear) draw range.

robertnurnberg:
> The model assumes symmetry indeed, now the data shows there is asymmetry (see top right graph near move 20), and trying to fit that is of course not possible.

Maybe a stupid question, but could we force the fitting also to be symmetrical? Essentially, replacing any tuple with a negative tuple with the "symmetricly" corresponding tuple? So just flipping eval in case of `D` and flipping eval and `L`<->`W` otherwise?

Alternatively, we could achieve the same affect by only keeping tuples for `D` and `W`, i.e. all `L` tuples in the current data would be replaced by the flipped `W` tuple.

vondele:
we could 'augment' the data to be symmetric, but there is no good reason to do it, IMO. Better to see the actual data, it could be the data is asymmetric for unknown reasons, which would merit investigation. I think what we see right now is more or less understood.

PR #47, opened 2023-09-26
let make detect changes to scoreWDLstat.cpp

robertnurnberg:


vondele:
the binary needs to have the dependencies on the .hpp in external/ as well, so it gets recompiled if those change

robertnurnberg:
Thanks, and done.

PR #49, opened 2023-09-29
add options --SPRTonly and --allowDuplicates

robertnurnberg:
In this PR we load the metadata centrally, and then use it to filer the list of pgn files based on different criteria:

1. If `--matchBook` is given, we only allow pgn files from tests that match the given regex (or do not match it, if `--matchBookInvert` is given). Observe that the logic compared to master was improved slightly: now any test without metadata or without the `"book"` key is rejected, if filtering based on book name is requested by the user.
2.  If `--SPRTonly` is given, we allow only pgn files from tests that have the `"sprt"` key in their metadata.
 
Once this is merged, I will finally add functionality to read the book depth from the metadata and adjust the full move counter appropriately.

In addition, this PR for the first time detects duplicate tests (in different directories) and warns the user about it. If `--allowDuplicates` is given, the analysis proceeds as in master. Otherwise the code exits.

### DIFF
--- a/scoreWDLstat.cpp
+++ b/scoreWDLstat.cpp
@@ -278,8 +278,8 @@ using map_meta = std::map<std::string, json>;
             test_map[test_id] = test_filename;
         } else if (test_map[test_id] != test_filename) {
             if (test_warned.find(test_filename) == test_warned.end()) {
-                std::cout << (allow_duplicates ? "Warning," : "Error:")
-                          << " detected a duplicate of test " << test_id << " in directory "
+                std::cout << (allow_duplicates ? "Warning" : "Error")
+                          << ": Detected a duplicate of test " << test_id << " in directory "
                           << directory << std::endl;
                 test_warned.insert(test_filename);
                 if (!allow_duplicates) {
--- a/scoreWDLstat.cpp
+++ b/scoreWDLstat.cpp
@@ -6,6 +6,7 @@
 #include <iostream>
 #include <mutex>
 #include <regex>
+#include <set>
 #include <string>
 #include <unordered_map>
 #include <vector>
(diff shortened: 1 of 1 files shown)


scoreWDLstat.cpp
```diff
@@ -258,33 +259,84 @@ void ana_files(map_t &map, const std::vector<std::string> &files, const std::str
...
+
+        if (test_map.find(test_id) == test_map.end()) {
+            test_map[test_id] = test_filename;
+        } else if (test_map[test_id] != test_filename) {
+            if (test_warned.find(test_filename) == test_warned.end()) {
+                std::cout << "Warning, detected a duplicate of test " << test_id << " in directory "
+                          << directory << std::endl;
+                test_warned.insert(test_filename);
+                if (!allow_duplicates) {
+                    std::cout << "Use --allowDuplicates to ignore this warning." << std::endl;
```
Disservin, line 285: `warning` but the program aborts ? Doesnt seem right

scoreWDLstat.cpp
```diff
@@ -258,33 +259,84 @@ void ana_files(map_t &map, const std::vector<std::string> &files, const std::str
...
 
-void filter_files(std::vector<std::string> &file_list, const std::regex &regex, bool invert) {
+void filter_files_book(std::vector<std::string> &file_list, const map_meta &meta_map,
+                       const std::regex &regex_book, bool invert) {
     file_list.erase(std::remove_if(file_list.begin(), file_list.end(),
-                                   [&regex, invert](const std::string &pgn_filename) {
-                                       std::string json_filename =
-                                           pgn_filename.substr(0, pgn_filename.find_last_of('-')) +
-                                           ".json";
+                                   [&regex_book, invert, &meta_map](const std::string &pathname) {
```
Disservin, line 308: those lambdas are pretty horrible to read, could you just move them into a variable perhaps?

Disservin:
Just had a quick look at the code, also `get_metadata` is pretty deeply nested, would be nice if there's some other "nicer" way, without so many nested if's.

robertnurnberg:
> Just had a quick look at the code, also `get_metadata` is pretty deeply nested, would be nice if there's some other "nicer" way, without so many nested if's.

Some of the nesting looking badly came from `make format`. I would prefer to leave stylistic changes for later (and maybe to others.) OK?

PR #50, opened 2023-09-29
add --verbose option

robertnurnberg:
Allows to limit the output from the download script.

(master version is clogging up my inbox with all those http errors that the cronjobs encounter)

May need some input on convention what to output when. E.g. by default the PR still prints info, could change to have no output unless at least `-v` is given, but that would change behaviour compared to master quite a bit.


vondele:
One could of course run the script with output redirected from within the crontab, but I'm fine with this functionality. It should still print some output though, I think this is probably good.

Issue #51, opened 2023-09-30
License


eduherminio:
It'd be great to explicitly state the license of this repository.

PR #52, opened 2023-09-30
add license

robertnurnberg:
Same license as SF.

Closes https://github.com/official-stockfish/WDL_model/issues/51

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,674 @@
+                    GNU GENERAL PUBLIC LICENSE
+                       Version 3, 29 June 2007
+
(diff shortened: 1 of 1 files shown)


vondele:
That would need the agreement of all code contributors. I'm fine, you as well, and so finally @Disservin should agree as well. There is no problem leaving the external code under MIT, that's compatible.

There are minor contributions from @dede1751  and @robbai that are probably not copyrightable, but I'd appreciate if they can OK this as well.

Disservin:
I'm fine with that as well.

vondele:
So I can merge that 
image

PR #58, opened 2023-10-03
add type annotations

Disservin:
- adds a few type annotations, some stuff is still `any`, where im a bit unsure currently/skipped for now

I think both @vondele and @robertnurnberg need to confirm that in their setup python is able to interpret this. The required minimum version should be `3.9` according to [vermin](https://github.com/netromdk/vermin) and if it's feasible to update their version if needed.


vondele:
On the system where it matters, I'm on 3.10.

robertnurnberg:
LGTM

Disservin:
Alright, thanks ;)

PR #60, opened 2023-10-04
simplify meta-data usage in cpp

Disservin:
- directly parse the json information into the struct

Code will fail, when the json is missing a key, can that be the case?

### DIFF
--- a/scoreWDLstat.cpp
+++ b/scoreWDLstat.cpp
@@ -175,8 +175,7 @@ void ana_files(map_t &map, const std::vector<std::string> &files, const std::str
             }
 
             if (meta_map.at(test_filename).book_depth.has_value()) {
-                auto book_depth = meta_map.at(test_filename).book_depth.value();
-                move_counter    = std::to_string(book_depth + 1);
+                move_counter = std::to_string(meta_map.at(test_filename).book_depth.value() + 1);
             } else {
                 if (!meta_map.at(test_filename).book.has_value()) {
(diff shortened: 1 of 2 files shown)


robertnurnberg:
Yes. See usage of `--onlySPRT` switch.

Disservin:
Mh true, I'll try to make it work with optional keys.

vondele:
this one will need a rebase

robertnurnberg:
Did you check for non-functionality? ;)

I get a nearly emtpy `updateWDL.json` file for the update script, locally.

scoreWDLstat.cpp
```diff
@@ -175,16 +174,17 @@ void ana_files(map_t &map, const std::vector<std::string> &files, const std::str
                 std::exit(1);
             }
 
-            if (meta_map.at(test_filename).find("book_depth") != meta_map.at(test_filename).end()) {
-                std::string book_depth = meta_map.at(test_filename)["book_depth"];
-                move_counter           = std::to_string(std::stoi(book_depth) + 1);
+            if (meta_map.at(test_filename).book_depth.has_value()) {
+                auto book_depth = meta_map.at(test_filename).book_depth.value();
```
robertnurnberg, line 178: I did this over two lines in master, to avoid having to cast to string in the second line. Can now be put into line I think.

Disservin:
> Did you check for non-functionality? ;)
> 
> I get a nearly emtpy `updateWDL.json` file for the update script, locally.

Mh let me check..

PR #61, opened 2023-10-04
simplify help command

Disservin:


scoreWDLstat.cpp
```diff
@@ -487,21 +488,25 @@ bool find_argument(const std::vector<std::string> &args,
...
-    std::cout << "  --matchBook <regex>   Filter data based on book name" << std::endl;
-    std::cout << "  --matchBookInvert     Invert the filter" << std::endl;
-    std::cout << "  --SPRTonly            Analyse only pgns from SPRT tests" << std::endl;
-    std::cout << "  --fixFEN              Patch move counters lost by cutechess-cli" << std::endl;
-    std::cout << "  -o <path>             Path to output json file (default: scoreWDLstat.json)"
-              << std::endl;
+    std::stringstream ss;
+
+    // clang-format off
+    ss << "Usage: " << program_name << " [options]" << "\n";
```
robertnurnberg, line 494: if we drop `std::endl` here (not sure if we should), we could move `\n` into the preceeding string?

Disservin: i dont really see a reason for either :p

PR #62, opened 2023-10-04
split source

Disservin:
- move some pure functions and struct's into a corresponding header

I think that way the "core" logic is a bit more easy to view/understand in `scoreWDLstat.cpp`.


vondele:
needs Makefile changes (add header dependency)

PR #66, opened 2023-10-06
allow parsing of .pgn.gz files

robertnurnberg:
Draft for now, may need some testing.

A future refactoring could try to drop the external lib.

In theory the local collections of .pgn's we each have can now be converted to .pgn.gz.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,7 +25,7 @@ HEADERS = external/chess.hpp external/json.hpp external/threadpool.hpp scoreWDLs
 all: $(EXE_FILE)
 
 $(EXE_FILE): $(SRC_FILE) $(HEADERS)
-	$(CXX) $(CXXFLAGS) $(NATIVE) -o $(EXE_FILE) $(SRC_FILE)
+	$(CXX) $(CXXFLAGS) $(NATIVE) -o $(EXE_FILE) $(SRC_FILE) -lz
 
 format:
 	clang-format -i $(SRC_FILE)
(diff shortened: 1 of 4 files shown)


robertnurnberg:
This needs https://github.com/official-stockfish/WDL_model/pull/65 to compile.

Disservin:
i was running into issues on windows, with this approach not reading the entire file... this more simple version works fine for me https://github.com/Disservin/WLD_model/tree/gzip-simple, performance can probably be improved

robertnurnberg:
Excellent, thanks. A much simpler solution. A quick test for me indicates it works as intended.

Will undraft for review.

robertnurnberg:
Undrafted. I have kept the commit history intact for now, but probably best to squash before merging. So let me know when I should squash.

Disservin (approved):
LGTM

Tested on windows and ubuntu.

robertnurnberg:
Thanks. Squashed into a single commit.

Disservin:
> Thanks. Squashed into a single commit.

wasnt really required anyway, since we can merge with squash? (and keep the correct co authorship)

robertnurnberg:
Oh, sorry. That wasn't my intention at all. If there is any way to restore your co-authorship, please go ahead.

In fact I thought this PR was a very nice team effort. 👍

Disservin:
I know ;) all good I don’t mourn the missing co authorship ^^

vondele:
> i was running into issues on windows, with this approach not reading the entire file... this more simple version works fine for me https://github.com/Disservin/WLD_model/tree/gzip-simple, performance can probably be improved

actually, I'm running in an issue with this trying to analyze the lichess db... a single file is 30GB compressed, so no way I can keep the full uncompressed file in memory. This would be a use case where reading from the file incrementally would be beneficial.

Disservin:
Oh well, I imagined something like this happen...

PR #67, opened 2023-10-07
detect gzipped downloads

robertnurnberg:
Correctly deal with gzip served data. Needs testing.

### DIFF
--- a/download_fishtest_pgns.py
+++ b/download_fishtest_pgns.py
@@ -1,6 +1,10 @@
 import urllib.request, urllib.error, urllib.parse
 import argparse, time, re, os, json
 
+def is_gz_file(filename):
+    with open(filename, 'rb') as f:
+        return f.read(2) == b'\x1f\x8b' 
+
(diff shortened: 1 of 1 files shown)


vondele:
Just wondering if we can't use the content header for this instead of looking at the magic bytes:
`if response.info().get('Content-Encoding') == 'gzip':`

from https://stackoverflow.com/questions/3947120/does-python-urllib2-automatically-uncompress-gzip-data-fetched-from-webpage

robertnurnberg:
I just thought with current fishtest we could simply rename the tmp file to `.pgn.gz` ? Without any checks. What do you think?

I guess we'll retire this code anyway once bulk download is implemented, so not worth it to spend more time on it?

vondele:
I agree we should retire this after bulk download. So kind of check is not bad. So if you agree, let's just merge that, to be removed as soon as bulk download arrives.

robertnurnberg:
yeah, just merge as is, thanks

PR #69, opened 2023-10-07
use API for metadata

robertnurnberg:
Draft for now, as it only gets the new metadata from the fishtest API.

~~I have not tested yet if this detect SPSA tests correctly (probably does not, will fix.)~~

But I wanted to publish this draft as it will need help from @Disservin to filter the json data within our .hpp file.
Could you check if the `"args"` key is present in the data, and then use the new format? Otherwise try the old format? Thanks a lot. You can push your changes to the .hpp file directly here.

### DIFF
--- a/scoreWDLstat.hpp
+++ b/scoreWDLstat.hpp
@@ -42,22 +42,9 @@ struct std::hash<Key> {
 };
 
 struct TestMetaData {
-    std::optional<std::string> base_net;
-    std::optional<std::string> base_options;
-    std::optional<std::string> base_tag;
     std::optional<std::string> book;
-    std::optional<std::string> last_updated;
-    std::optional<std::string> new_net;
-    std::optional<std::string> new_options;
-    std::optional<std::string> new_tag;
-    std::optional<std::string> new_tc;
     std::optional<std::string> sprt;
-    std::optional<std::string> start_time;
-    std::optional<std::string> tc;
-
-    std::optional<int> threads;
     std::optional<int> book_depth;
-    std::optional<bool> adjudication;
 };
 
 template <typename T = std::string>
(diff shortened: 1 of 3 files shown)


robertnurnberg:
spsa detection fixed and tested. seems to work.

Disservin:
i think this should do it? please test ^^

robertnurnberg:
wow, that was quick. thanks a lot. will be afk now, you could test the update script for master and patch for same .png output. will undraft just in case.

scoreWDLstat.hpp
```diff
@@ -60,41 +60,62 @@ struct TestMetaData {
...
+    auto j = nlohmann_json_j;
+
+    // Check if args key is present, if so we are using metadata from the fishtest api
+    if (j.find("args") != j.end()) {
+        nlohmann_json_t.start_time   = get_optional(nlohmann_json_j, "start_time");
+        nlohmann_json_t.last_updated = get_optional(nlohmann_json_j, "last_updated");
+
+        j = j["args"];
+
+        nlohmann_json_t.adjudication = get_optional<bool>(j, "adjudication");
```
robertnurnberg, line 83: should this be made consistent with the treatment below?

scoreWDLstat.hpp
```diff
@@ -60,41 +60,62 @@ struct TestMetaData {
...
+
+        nlohmann_json_t.threads = get_optional<int>(j, "threads");
+
+        nlohmann_json_t.book_depth =
+            get_optional(j, "book_depth").has_value()
+                ? std::optional<int>(std::stoi(get_optional(j, "book_depth").value()))
```
robertnurnberg, line 89: I am wondering why `"book_depth"` returns a string, and not an int like other fields. I can only presume that for non-uniform book depths, this field still contains some info, like `"variable 4-8"` or similar? Maybe @vondele can confirm? 
In that case, maybe we should catch here (and below) if the string contains more than just a number.

scoreWDLstat.hpp
```diff
@@ -60,41 +60,62 @@ struct TestMetaData {
...
+        nlohmann_json_t.threads =
+            get_optional(j, "threads").has_value()
+                ? std::optional<int>(std::stoi(get_optional(j, "threads").value()))
+                : std::nullopt;
+    }
+
+    nlohmann_json_t.tc           = get_optional(j, "tc");
+    nlohmann_json_t.base_net     = get_optional(j, "base_net");
+    nlohmann_json_t.base_options = get_optional(j, "base_options");
+    nlohmann_json_t.base_tag     = get_optional(j, "base_tag");
```
robertnurnberg, line 113: just a warning that the contents of `base_tag` and `new_tag`has changed now. not sure how best to proceed. we could create the new `resolved_base` and `resolved_new` and extract that info from the old style metadata in `base_tag` etc. Or we drop these entries completely for now, and only introduce them later, when we need them. And then only do so for the new format. up to you. :)

scoreWDLstat.hpp
robertnurnberg: PS: Could the book_depth extraction be merged for the two if branches?

robertnurnberg:
> i think this should do it? please test ^^

I tested this with new and old metadata, with intact and broken metadata. All works as expected. :+1:

scoreWDLstat.hpp
vondele: this book_depth, most probably can only be an int (i.e. fishtest should be passing an int to cutechess). Now, note that this is the maximum depth the book moves should be played for pgn books, so that's not quite what we would use to fixup the actual move number in pgns.

robertnurnberg: Agreed. This only works as intended in conjunction with a filter on book name.

Regarding int or string: we may want to make this future proof by first checking if it's an int, and only if not convert from string. This way we are protected from future changes on the fishtest side.

Disservin:
Okay, I'm inclined to simply through away all meta data which we dont use for now. 

Regarding the book depth I think it's an issue in fishtest, the worker also always casts it to int https://github.com/official-stockfish/fishtest/blob/master/worker/games.py#L1398.
Typically the usage with the json lib is way easier since you can just use `NLOHMANN_DEFINE_TYPE_NON_INTRUSIVE` however this assumes the types are correct and present, I havent see a similar macro which does the optional type definitions..

I am not sure why the book_depth is a string, when it comes back from the api, since in fishtest it's always treated as number/int (and always casted too) cc @ppigazzini

vondele:
I guess this one is ready to go, right?

robertnurnberg:
I think in light of https://github.com/official-stockfish/WDL_model/pull/77 this can probably be simplified: we should only deal with the latest version of the metadata and reject (with error) or ignore incomplete data.

Disservin:
Yeah this still needs some work

Disservin:
Okay I just pushed the simplification, which removes unused keys and now only works for the new json format.

robertnurnberg:
Ok, this needs to be carefully tested. Testing master with master metadata, then patch with new metadata (ideally without losing the old metadata in case things don't work out.) Will let you know.

robertnurnberg:
Hm, first test with new metadata and patch gives me an empty `updateWDL.json`.

Disservin:
ah i left out sprt

Disservin:
added the sprt key

Disservin:
okay, please test again ;)

robertnurnberg:
```
> diff master.log patch.log
4,6c4,6
Progress: 96/96
< Time taken: 98.845s
< Wrote 820590913 scored positions to updateWDL.json for analysis.
---
Progress: 96/96
> Time taken: 164s
> Wrote 820601266 scored positions to updateWDL.json for analysis.
```

So mismatch in position count. I will be afk now, but you have my pgn selection to reproduce. Basically I ran
`find . -name "*.json" -print | zip metadata -@` to save the old metadata, then used `download_missing_metadata.py` to download the new one.

In theory there could have been old pgns without metadata that cause the increase in positions now? Would need some time to check this. Sadly, will be travelling tomorrow....

robertnurnberg:
Rebased on master. Converted to draft because I would like to understand what caused the diff before we merge this. E.g. patch could now also be compared to commit mentioned in https://github.com/official-stockfish/WDL_model/pull/80#issuecomment-1761691365, which for me was non-functional with the chess-lib changes.

robertnurnberg:
All good now. For me rebased PR is equivalent to master (with old metadata) is equivalent to 57c7befa748c976a6ea9060f95481e3999c23e5f (with old metadata). 

So from my point of view this is ready to be merged.

PR #72, opened 2023-10-09
gzip stream

Disservin:
i think this should avoid reading the entire file into memory

PR #75, opened 2023-10-09
update chess-library

Disservin:
- should fix the castling parsing bug encountered with #72


Disservin:
This should fix the parsing of castling moves with annotations. If the pgn parser is completely capable to parse all kinds of lichess pgns is I think still to be determined.

PR #77, opened 2023-10-12
add script to download missing metadata for existing pgn data

robertnurnberg:
The new script allows to download the new(!) metadata for any test that pgn files can be found for locally. With the option `-o` it can be used to override existing metadata. (To be used with care, and certainly not before https://github.com/official-stockfish/WDL_model/pull/69 is merged.)

Not quite sure whether this should be added to the repo, but the script is quite useful I think.

- It means that the PR https://github.com/official-stockfish/WDL_model/pull/69 can be simplified and support of legacy metadata can be dropped completely.
- It means that @vondele can in future also filter very old pgn data to not include FRC data, or to only use a UHO book, in connection with his https://github.com/vondele/fastpopular.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,7 +29,7 @@ $(EXE_FILE): $(SRC_FILE) $(HEADERS)
 
 format:
 	clang-format -i $(SRC_FILE)
-	black -q download_fishtest_pgns.py scoreWDL.py
+	black -q download_fishtest_pgns.py scoreWDL.py download_missing_metadata.py
 	shfmt -w -i 4 updateWDL.sh
 
 clean:
--- a/download_missing_metadata.py
+++ b/download_missing_metadata.py
@@ -1,0 +1,55 @@
+import argparse, json, os, re, requests, tarfile, time, urllib.request
+
+parser = argparse.ArgumentParser(
+    description="Download fishtest metadata for any test that .pgn(.gz) files can be found for.",
+    formatter_class=argparse.ArgumentDefaultsHelpFormatter,
+)
+
+parser.add_argument(
+    "--path",
+    default="./pgns",
+    help="Recursively look for .pgn(.gz) files in this directory.",
+)
+parser.add_argument(
+    "-o",
+    "--overwrite",
+    action="store_true",
+    help="Overwrite existing metadata (use with care!).",
+)
+args = parser.parse_args()
+
+if not os.path.exists(args.path):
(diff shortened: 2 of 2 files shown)


robertnurnberg:
Similarly to the first point, this will make it easy to react to changes in the fishtest api. For example, if `"book_depth"` becomes an `int` in future, we can simply adapt the cpp code to switch to `int` and expect users to update their metadata with the script from this PR.

vondele:
so, that's certainly something useful, I think I need to start moving my old collection of pgns (which are basically just directories containing plenty of pgns of all tests together) to the current directory structure, and gzipping it on the fly, and updating it with metadata...

robertnurnberg:
No need to restructure the directories, unless you really want to: the script (and our .cpp analyser, as well as your fastpopular spin-off) will accept different metafiles in the same dir for different tests. I think!

robertnurnberg:
> so, that's certainly something useful, I think I need to start moving my old collection of pgns (which are basically just directories containing plenty of pgns of all tests together) to the current directory structure, and gzipping it on the fly, and updating it with metadata...

The line I used to gzip my collection was `find . -type f -name "*.pgn" -exec gzip {} \;`, btw.

vondele:
And I used `find . -type f -name '*.pgn' -print0 | parallel -0 -j 16 gzip {}`  to do it in parallel.

vondele:
Found 5996 downloaded tests in /mnt/md0/chess/fishtestgames/. ....

vondele:
I'm seeing a couple of 
`  error: caught exception "Expecting value: line 1 column 1 (char 0)" `
in the script output.

vondele:
Ok, that's the real exception:
`error: /mnt/md0/chess/fishtestgames/download33/61efcf3b977220a27e32612c.json : caught exception "502 Server Error: Bad Gateway for url: https://tests.stockfishchess.org/api/get_run/61efcf3b977220a27e32612c"` it basically is a temporary server error.

robertnurnberg:
Hm, I guess you can manually delete the corrupted json files with a `find` one liner and restart the script?

vondele:
Improved the error handling e0c128af04e2d905f64fc2c231f3b9890b2ac2fa

vondele:
yeah, just rerunning the script eventually gets all json. I don't think it leads to corrupted json files, so that's OK.

PR #81, opened 2023-10-14
skip tests that do not match revisions

robertnurnberg:
Use the metadata to skip whole tests if they do not involve the SF revisions of interest.

`> diff master.log patch.log`:
```diff
1a2
> Found 101774 .pgn(.gz) files in total.
3c4,5
< Found 73747 .pgn(.gz) files, creating 96 chunks for processing.
---
> Filtering pgn files matching revision SHA .*38e830af4bfa6c9e9c11279a8e6a60b6ca4ec2cd|.*002636362e175134c6d0d53b332b527ec4a12db0|.*7a4de96159f76f2465d474d76e08a1c8ca3383b8|.*f7fbc6880efae2ec9d97d6f1d65e2ad00547e32c|.*25d444ed60e3873c02a70525776b145f03833103|.*008d59512ac38e1e4a2f7880fe4e07b902845bb0|.*c17a657b045d4dc720c8c36558fe649a1c3f4a05|.*040dfedb3457ca6971d98c754362cde4dc767aff|.*f1ce1cd4751a098b7ee09e304fa6397d08fe8d7f|.*8a912951de6d4bff78d3ff5258213a0c7e6f494e|.*afe7f4d9b0c5e1a1aa224484d2cd9e04c7f099b9|.*660da1ca7b4c2c03dce03d14ef3496d9fb4aead2|.*4f0fecad8a0f5258114f63f0ac0c905a54d65219|.*31d0b7fe932458d6661f4d4c2ce88502086616c5|.*243f7b264a81c2981cec2818b47d609d9d3ca119|.*9739ed7a97c153c3223b608b24717edbf2dfd7bc|.*ce99b4b2ef74d09499f35f09bc33102d203791cd|.*22cdb6c1ea1f5ca429333bcbe26706c8b4dd38d7|.*70ba9de85cddc5460b1ec53e0a99bee271e26ece
> Found 48737 .pgn(.gz) files, creating 96 chunks for processing.
5c7
< Time taken: 117.269s
---
> Time taken: 79.278s
```

A pitfall for future users may be that they devise a regex for the commit SHA, which then does not work for the engine name in the pgn files. I put extra info in help output, but still something to be aware of. I  do not think there is an elegant and easy solution to resolve this. At the end of the day, users will be best advised to use separate regexes for revisions and engine names.

### DIFF
--- a/scoreWDLstat.cpp
+++ b/scoreWDLstat.cpp
@@ -344,6 +344,28 @@ void filter_files_book(std::vector<std::string> &file_list, const map_meta &meta
     file_list.erase(std::remove_if(file_list.begin(), file_list.end(), pred), file_list.end());
 }
 
+void filter_files_revision(std::vector<std::string> &file_list, const map_meta &meta_map,
+                           const std::regex &regex_rev) {
+    const auto pred = [&regex_rev, &meta_map](const std::string &pathname) {
+        std::string test_filename = pathname.substr(0, pathname.find_last_of('-'));
+
+        if (meta_map.find(test_filename) == meta_map.end()) {
+            return true;
+        }
+        if (meta_map.at(test_filename).resolved_base.has_value() &&
+            std::regex_match(meta_map.at(test_filename).resolved_base.value(), regex_rev)) {
+            return false;
+        }
+        if (meta_map.at(test_filename).resolved_new.has_value() &&
+            std::regex_match(meta_map.at(test_filename).resolved_new.value(), regex_rev)) {
+            return false;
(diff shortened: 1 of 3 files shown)


vondele:
Ideally we factor out a bit the functionality related to filtering so it can be easily shared between this repo and fastpopular. That part of the functionality is very similar.

robertnurnberg:
Agreed, but I'd probably leave that to you and disservin. One decision is whether to have a single filter function, or three separate ones as now....

PR #83, opened 2023-10-15
output total number of games analysed

robertnurnberg:
I think the total number of games is nice to have, e.g. for the commit message of WDL PRs to stockfish.

`> diff master.log patch.log` 
```diff
7,8c7,8
< Time taken: 77.598s
< Wrote 916508997 scored positions to updateWDL.json for analysis.
---
> Time taken: 78.934s
> Wrote 916508997 scored positions from 120116780 games to updateWDL.json for analysis.
```

### DIFF
--- a/scoreWDLstat.cpp
+++ b/scoreWDLstat.cpp
@@ -36,6 +36,7 @@ map_t pos_map = {};
 using map_meta = std::unordered_map<std::string, TestMetaData>;
 
 std::atomic<std::size_t> total_chunks = 0;
+std::atomic<std::size_t> total_games  = 0;
 
 namespace analysis {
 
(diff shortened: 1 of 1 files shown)


robertnurnberg:
@vondele I am thinking with 120M games and nearly 1G positions, soon might be a good time to update SF's WDL model. My main reason would be to finally correct the 8-move offset. Would you mind if I open such a PR some time next week?

vondele:
you can certainly make the next PR for that... however, it would probably make sense to have the data start at the latest net architecture, and currently there are some tests running like https://tests.stockfishchess.org/tests/view/652b3a23de6d262d08d2d63a that actually scale the NNUE eval quite a bit, which directly influences the eval shown. If one of these eval tunes passes and gets merged that would be the logical point to start collecting 'production' data.

robertnurnberg:
Ok, thanks. I will keep an eye on the SF commits and change the `firstrev` in `updateWDL.sh` once net arch or eval scaling changes, to reset data collection. For now `firstrev` is still current in terms of most recent net arch change, right?

PR #84, opened 2023-10-19
update chess library

robertnurnberg:
This fixes two possible issues in pgn move parsing.

### DIFF
--- a/scoreWDLstat.cpp
+++ b/scoreWDLstat.cpp
@@ -195,7 +195,7 @@ class Analyze : public pgn::Visitor {
 
     void endPgn() override {
         board.set960(false);
-        board.setFen(STARTPOS);
+        board.setFen(chess::constants::STARTPOS);
 
         goodTermination = true;
         hasResult       = false;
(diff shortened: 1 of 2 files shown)


scoreWDLstat.cpp
```diff
@@ -195,7 +195,7 @@ class Analyze : public pgn::Visitor {
 
     void endPgn() override {
         board.set960(false);
-        board.setFen(STARTPOS);
+        board.setFen(chess::constants::STARTPOS);
```
Disservin, line 198: no need for the `chess` prefix

Disservin:
next would be also to make use of the `skipPgn(true)` functionality, allows for a bit easier pgn skipping

robertnurnberg:
The pgn parsing from within the chess-library should be fixed now.

Disservin:
@robertnurnberg I get the same scored position with these changes, you too?

As a side note it seems the parser is slightly slower at least for me? Not of much importance anyway.
```
master: Time taken: 8.539s
patch:  Time taken: 10.886s
```

robertnurnberg:
Yes, I tested for non-functionality on a (small) test set of pgns.

PR #87, opened 2023-10-22
fix types and --no-fit

Disservin:
- better python types
- fix `--no-fit`, this previously threw `AttributeError: 'tuple' object has no attribute 'model_ms'`
      - I still get `No artists with labels found to put in legend.  Note that artists whose label start with an underscore are ignored when legend() is called with no argument.`... not sure why maybe someone has an idea

### DIFF
--- a/scoreWDL.py
+++ b/scoreWDL.py
@@ -347,27 +347,32 @@ class WdlModel:
 
         fit = ModelFit(self.args.yDataTarget, self.args.NormalizeToPawnValue)
 
-        if self.args.fit and model is not None:
-            self.plot.axs[1, 0].plot(model.model_ms, model.model_as, "b.", label="as")
-            self.plot.axs[1, 0].plot(
-                model.model_ms,
-                fit.poly3(model.model_ms, *model.popt_as),
-                "r-",
-                label="fit: " + model.label_as,
-            )
-            self.plot.axs[1, 0].plot(model.model_ms, model.model_bs, "g.", label="bs")
-            self.plot.axs[1, 0].plot(
-                model.model_ms,
-                fit.poly3(model.model_ms, *model.popt_bs),
-                "m-",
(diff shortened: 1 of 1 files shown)


robertnurnberg:
The error (which is harmless) is caused by the two `legend` calls, as these plots are not really used.

If you want to get rid of the error message, just call legend only in the non `--no-fit` cases. Best practice might be to only refer to the two subplot `[0][1]` and `[0][2]` in the `--no-fit` case, i.e. don't change the others at all.

Disservin:
> The error (which is harmless) is caused by the two `legend` calls, as these plots are not really used.
> 
> If you want to get rid of the error message, just call legend only in the non `--no-fit` cases. Best practice might be to only refer to the two subplot `[0][1]` and `[0][2]` in the `--no-fit` case, i.e. don't change the others at all.

you mean like this, regarding the easy fix :D ?
```diff
diff --git a/scoreWDL.py b/scoreWDL.py
index 7e5deba..bd9efd4 100644
--- a/scoreWDL.py
+++ b/scoreWDL.py
@@ -365,7 +365,10 @@ class WdlModel:

         self.plot.axs[1, 0].set_xlabel(self.args.yData)
         self.plot.axs[1, 0].set_ylabel("parameters (in internal value units)")
-        self.plot.axs[1, 0].legend(fontsize="x-small")
+
+        if self.args.fit is None:
+            self.plot.axs[1, 0].legend(fontsize="x-small")
+
         self.plot.axs[1, 0].set_title("Winrate model parameters")
         self.plot.axs[1, 0].set_ylim(bottom=0.0)
         ```

robertnurnberg:
but better to bracket the whole `[1,0]` code

Disservin:
okay changed so that 1,0 is only run for fits... small overlap now with #97

PR #89, opened 2023-10-23
Adapt pgn downloader to new fishtest API's

peregrineshahin:
some simplifications and added some options.

### DIFF
--- a/download_fishtest_pgns.py
+++ b/download_fishtest_pgns.py
@@ -1,7 +1,7 @@
-import argparse, json, os, re, requests, tarfile, urllib.request
+import argparse, datetime, json, os, re, requests, tarfile, urllib.request
 
 parser = argparse.ArgumentParser(
-    description="Bulk-download .pgn.gz files from completed LTC tests on fishtest.",
+    description="Bulk-download .pgn.gz files from finished tests on fishtest.",
     formatter_class=argparse.ArgumentDefaultsHelpFormatter,
 )
 
@@ -11,10 +11,34 @@ parser.add_argument(
     help="Downloaded .pgn.gz files will be stored in PATH/date/test-id/.",
 )
 parser.add_argument(
-    "--page",
-    type=int,
-    default=1,
-    help="Page of completed LTC tests to download from.",
+    "--ltc_only",
+    type=bool,
+    default=True,
+    help="A flag for LTC tests only.",
+)
+parser.add_argument(
+    "--time_delta",
+    type=float,
+    default=168.0,  # 1 week
+    help="Delta of hours from now since the desired tests are last updated.",
+)
+parser.add_argument(
+    "--success_only",
+    type=bool,
+    default=False,
+    help="A flag for Green tests only.",
+)
+parser.add_argument(
+    "--yellow_only",
+    type=bool,
+    default=False,
+    help="A flag for yellow tests only.",
+)
+parser.add_argument(
+    "--username",
+    type=str,
+    default="",
+    help="Specified username to download from their finished tests.",
 )
 parser.add_argument(
     "-v",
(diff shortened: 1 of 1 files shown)


download_fishtest_pgns.py
```diff
@@ -89,27 +141,25 @@
...
+            if args.verbose >= 2:
+                print("Extracting the tar file ...")
+            with tarfile.open(tmpName, "r") as tar:
+                tar.extractall(path)
+            os.remove(tmpName)
+        except Exception as ex:
+            if args.verbose >= 2:
+                print(f'  error: caught exception "{ex}"')
+            continue
+    page += 1  # Move to the next page for the next iteration
```
robertnurnberg, line 165: so I am not sure about this `page+=1`. That means by default the script goes through all the pages? Meaning it would download everything that is available?
would need feedback from @vondele 
but if we go via that route, then the command line option `--page` can probably be removed

robertnurnberg: hm, in general might create quite a bit of unnecessary requests to the server? (how many pages are there?)

so maybe just looking at specified page is enough, user can always call the scripts several times with different `--page` options ...

robertnurnberg: (as discussed on discord, nice solution might be to hide page aspect completely from user, let user give time span (in hours or days), and compute needed unix time stamp from now)

robertnurnberg:
I get `AttributeError: type object 'datetime.datetime' has no attribute 'datetime'`.

robertnurnberg:
And I think the `page` aspect is best hidden completely from the user, as this is an implementation detail: let user simply specify the time delta, and the script then gets _all_ the tests available within that time frame.

robertnurnberg:
did a quick test with `python download_fishtest_pgns.py --time_delta 0` and would have thought nothing gets downloaded: but download of a test started

peregrineshahin:
> did a quick test with `python download_fishtest_pgns.py --time_delta 0` and would have thought nothing gets downloaded: but download of a test started
> 
> Also with `python download_fishtest_pgns.py --time_delta 1` surprisingly many tests get downloaded ;)

Please try now

download_fishtest_pgns.py
```diff
@@ -11,10 +11,34 @@
...
+    help="A flag for LTC tests only, coded values:[0, 1].",
+)
+parser.add_argument(
+    "--time_delta",
+    type=float,
+    default=168.0,  # 1 week
+    help="Delta of hours from now since the desired tests are last updated.",
+)
+parser.add_argument(
+    "--success_only",
```
vondele, line 26: as these are effectively bools, should they be that way for input, and converted to int just for the api call?

download_fishtest_pgns.py
```diff
@@ -11,10 +11,34 @@
     help="Downloaded .pgn.gz files will be stored in PATH/date/test-id/.",
 )
 parser.add_argument(
-    "--page",
+    "--ltc_only",
     type=int,
-    default=1,
-    help="Page of completed LTC tests to download from.",
+    default=0,
```
robertnurnberg, line 16: so current setup of the repo would need `--ltc_only` to be true by default

as the analysis code does not filter based on TC yet, I think for now this should stay

download_fishtest_pgns.py
```diff
@@ -11,10 +11,34 @@
...
-    "--page",
-    type=int,
-    default=1,
-    help="Page of completed LTC tests to download from.",
+    "--ltc_only",
+    type=bool,
+    default=True,
+    help="A flag for LTC tests only.",
+)
+parser.add_argument(
```
robertnurnberg, line 19: tiny nit: I would change the order of these two cli options, so `--time_delta` first (mainly for `--help` to have a more consistent order)

robertnurnberg:
LGTM, thanks a lot

PR #92, opened 2023-10-24
introduce --fixFENsource

robertnurnberg:
Nonfunctional change for me when running `./updateWDL.sh` .

Some comments:

- As indicated on [discord](https://discord.com/channels/435943710472011776/1032922913499783169/1166296059182579803), the book `UHO_4060_v3.epd` seems to contain impossible ep squares, which are then removed by cutechess-cli.
- Hence the `std::unordered_map<std::string, FixFenData>` uses the first _three_ FEN fields as key, not the first four.
- Storing the ep data from the book is not really necessary, and can be simplified away in future. Then `FixFenData` could just be replaced by `std::pair<int, int>`.
- Currently the update script can only deal with a single book. Some thought is needed how to allow multiple books. (download separately, create a merged `book.epd.gz`, or feed regex to the analysis script for books to read in for reference).
- But already in its current state it can be used to analyse games from only the new book.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,8 +3,10 @@
 *.pgn
 *.gz
 *.log
+*.exe
+
+scoreWDLstat
 
 pgns*
-scoreWDLstat
-*.exe
 Stockfish
+books
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,7 @@
 *.png
 *.json
 *.pgn
-*.pgn.gz
+*.gz
 *.log
 
 pgns*
--- a/scoreWDLstat.cpp
+++ b/scoreWDLstat.cpp
@@ -32,9 +32,21 @@ using map_t =
 
 map_t pos_map = {};
 
-// maps to collect metadata for tests, and move counters for fix FENs
+// map to collect metadata for tests
 using map_meta = std::unordered_map<std::string, TestMetaData>;
-using map_fens = std::unordered_map<std::string, std::pair<int, int>>;
+
+// class (and map) to hold data that cutechess-cli lost from original FENs
+class FixFenData {
+   public:
+    std::string ep;
+    int halfmove;
+    int fullmove;
+
+    FixFenData() {}
+    FixFenData(const std::string &ep, int halfmove, int fullmove)
+        : ep(ep), halfmove(halfmove), fullmove(fullmove) {}
+};
+using map_fens = std::unordered_map<std::string, FixFenData>;
 
 std::atomic<std::size_t> total_chunks = 0;
 std::atomic<std::size_t> total_games  = 0;
@@ -83,7 +95,7 @@ class Analyze : public pgn::Visitor {
 
     void header(std::string_view key, std::string_view value) override {
         if (key == "FEN") {
-            std::regex p("^(.+) 0 1$");
+            std::regex p("^(.+) - 0 1$");
             std::smatch match;
             std::string value_str(value);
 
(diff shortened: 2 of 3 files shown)


Disservin:
I can review later today, now I wonder what the performance cost of all that is, lot's of strings and a new regex.

robertnurnberg:
For me patch is faster. ;)

```
> diff scoreWDLstat.log master.log
6c6
Progress: 83/96Error when parsing: pgns/23-10-16/652d94dede6d262d08d30c99/652d94dede6d262d08d30c99-535.pgn.gz
---
Progress: 84/96Error when parsing: pgns/23-10-16/652d94dede6d262d08d30c99/652d94dede6d262d08d30c99-535.pgn.gz
8,9c8,9
Progress: 96/96
< Time taken: 845.324s
---
Progress: 96/96
> Time taken: 897.588s
```

robertnurnberg:
Just for reference, for

```diff
diff --git a/updateWDL.sh b/updateWDL.sh
index fdc6a46..934c671 100755
--- a/updateWDL.sh
+++ b/updateWDL.sh
@@ -10,7 +10,7 @@ firstrev=70ba9de85cddc5460b1ec53e0a99bee271e26ece
 lastrev=HEAD
 
 # "regex" for book name, TODO: only works for single book name for now!
-bookname=UHO_4060_v3.epd
+bookname=UHO_Lichess_4852_v1.epd
 
 # path for PGN files
 pgnpath=pgns
@@ -69,8 +69,8 @@ echo "Look recursively in directory $pgnpath for games from SPRT tests using" \
 ./scoreWDLstat --dir $pgnpath -r --matchRev $regex_pattern --matchBook "$bookname" --fixFENsource "$bookname.gz" --SPRTonly -o updateWDL.json >&scoreWDLstat.log
 
 # fit the new WDL model, keeping anchor at move 32
-# we ignore the first 2 full moves out of book for fitting (11=8+1+2), and the first 9 for (contour) plotting (18=8+1+9)
-python scoreWDL.py updateWDL.json --plot save --yDataTarget 32 --yDataMin 11 --yDataMax 120 --yPlotMin 18 --NormalizeToPawnValue $oldpawn >&scoreWDL.log
+# we ignore the first 2 full moves out of shortest book exit (2 plies) for fitting (4=1+1+2), and the first 2 for (contour) plotting (4=1+1+2)
+python scoreWDL.py updateWDL.json --plot save --yDataTarget 32 --yDataMin 4 --yDataMax 120 --yPlotMin 4 --NormalizeToPawnValue $oldpawn >&scoreWDL.log
 
 # extract the total number of positions, and the new NormalizeToPawnValue
 poscount=$(awk -F '[() ,]' '/Retained \(W,D,L\)/ {sum = 0; for (i = 9; i <= NF; i++) sum += $i; print sum; exit}' scoreWDL.log)
 ```
 
 so without hiding the empty contour plot regions, and fitting from move 4 onwards, I get this with `./updateWDL.sh`.
WDL_model_summary

vondele:
For me the patch is working, but I'll wait for an OK from Disservin.

Disservin:
Some stuff could probably be better written but it doesnt matter now,  can be refactored at some point in the future.

robertnurnberg:
> Some stuff could probably be better written but it doesnt matter now, can be refactored at some point in the future.

Agreed, and thanks for merging! :)

Btw, does the latest SF patch mean we should rebase `updateWDL.sh` on that commit? (@vondele mention some eval rescaling in the past)

vondele:
no, that patch is still just search. Interestingly, linrock has many of these eval tweaks passing STC, but they are currently all failing LTC. https://tests.stockfishchess.org/tests/user/linrock

PR #95, opened 2023-10-26
ignore duplicate lines and lines without full move counter

robertnurnberg:
This PR fixes a small oversight in `scoreWDLstat.cpp` that would have led to corrupt data if the reference file contained EPDs without move counters.

In addition, when creating the reference file in `updateWDL.sh`, we now immediately reject lines without move counters, and we remove any duplicate lines. This allows us to include also tests with older UHO books in the analysis. (This latter point seems to be irrelevant for my downloaded tests.)

Incidentally, I noticed that even for the _same_ book, if the lines are sorted, then the resulting `.gz` is much smaller! Something to keep in mind for future fishtest books, maybe (assuming that same holds for the `.zip` files). @vondele 

`diff master.log scoreWDLstat.log`
```diff
3c3
< Filtering pgn files matching the book name UHO_4060_v3.epd|UHO_Lichess_4852_v1.epd
---
> Filtering pgn files matching the book name UHO_4060_v..epd|UHO_Lichess_4852_v1.epd
6c6
Progress: 64/96Error when parsing: pgns/23-10-16/652d94dede6d262d08d30c99/652d94dede6d262d08d30c99-535.pgn.gz
---
Progress: 63/96Error when parsing: pgns/23-10-16/652d94dede6d262d08d30c99/652d94dede6d262d08d30c99-535.pgn.gz
8,9c8,9
Progress: 96/96
< Time taken: 838.858s
---
Progress: 96/96
> Time taken: 834.52s
```

### DIFF
--- a/scoreWDLstat.cpp
+++ b/scoreWDLstat.cpp
@@ -286,12 +286,15 @@ void ana_files(const std::vector<std::string> &files, const std::string &regex_e
             int halfmove, fullmove = 0;
 
             iss >> f1 >> f2 >> f3 >> ep >> halfmove >> fullmove;
+
+            if (!fullmove) continue;
+
             key              = f1 + ' ' + f2 + ' ' + f3;
             auto fixfen_data = std::pair<int, int>(halfmove, fullmove);
 
             if (fixfen_map.find(key) != fixfen_map.end()) {
                 // for duplicate FENs, prefer the one with lower full move counter
-                if (fullmove && fullmove < fixfen_map[key].second) {
+                if (fullmove < fixfen_map[key].second) {
                     fixfen_map[key] = fixfen_data;
                 }
             } else {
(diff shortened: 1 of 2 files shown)


robertnurnberg:
Here the size difference old and new (with new _much_ smaller, even though it includes about the same number of lines)
```
46830485 Oct 25 19:18 fixfen_29fba70df70aaf571060ab14ad13ff8e.epd.gz
28550483 Oct 26 08:54 fixfen_8e179224ec8730b8db2639c2172787c4.epd.gz
```

```
> zcat fixfen_29fba70df70aaf571060ab14ad13ff8e.epd.gz | wc -l 
2874237
> zcat fixfen_8e179224ec8730b8db2639c2172787c4.epd.gz | wc -l
2871525
```

vondele:
yes, I've observed the sorted is compressed better before (not surprising though), however, there is some loss of 'meta-information' that is implicit. Now the book lines are sorted approximately in frequency order. Also, if people don't use the book randomized in cutechess, the starting positions would be very similar after a sort.

PR #97, opened 2023-10-28
Fix winmodel arg types, and add a TODO

vondele:


vondele:
This became just a comment / TODO after #87.

PR #98, opened 2023-10-29
Add a few comments, rename variable

vondele:


scoreWDL.py
```diff
@@ -40,6 +40,7 @@ def __init__(self, filenames: list[str]):
         self.filenames = filenames
 
     def load_json(self) -> dict[str, int]:
+        """Load the json, which describes with the key the position, and the value the count of these positions"""
```
robertnurnberg, line 43: does not read that well atm, maybe insert `with` before `the value`?

vondele: yeah, new attempt.

scoreWDL.py
```diff
@@ -40,6 +40,9 @@ def __init__(self, filenames: list[str]):
         self.filenames = filenames
 
     def load_json(self) -> dict[str, int]:
+        """Load the json, in which the key describes the position (result, move, material),
```
robertnurnberg, line 43: could make this consistent with readme: `outcome, move, material, eval`

vondele: right now it is consistent with the code a bit below, that refers to this as 'result'

vondele: ah, right.

robertnurnberg:
you could merge this and base the other PR on this one? (also exclude the `.png` change in the other one)

PR #100, opened 2023-11-02
Allow for optimizing the model based on better defined properties

vondele:
* fitDensity: original approach just fitting the win rate, and corresponding parameters
* optimizeProbability: maximimize the probability with which the model predicts the game outcome
* optimizeScore: minimize the square error between the actual game score, and the predicted game score.

### DIFF
--- a/scoreWDL.py
+++ b/scoreWDL.py
@@ -2,9 +2,10 @@ import json, argparse, numpy as np, matplotlib.pyplot as plt
 from collections import Counter
 from ast import literal_eval
 from scipy.interpolate import griddata
-from scipy.optimize import curve_fit
+from scipy.optimize import curve_fit, minimize
 from dataclasses import dataclass
 from typing import Literal, Callable, Any
+import math
 
 
 class WdlPlot:
@@ -27,12 +28,14 @@ class WdlPlot:
 
 
 @dataclass
-class RawModelData:
-    xs: list[float]
-    ys: list[int]
-    zwins: list[float]
-    zdraws: list[float]
-    zlosses: list[float]
+class ModelDataDensity:
+    """Count data converted to densities"""
+
+    xs: list[float]  # scores
+    ys: list[int]  # moves or material
+    zwins: list[float]  # corresponding win probability
+    zdraws: list[float]  # draw prob
+    zlosses: list[float]  # loss prob
 
 
 class DataLoader:
@@ -96,12 +99,12 @@ class DataLoader:
         )
         return win, draw, loss
 
-    def get_raw_model_data(
+    def get_model_data_density(
         self,
         win: Counter[tuple[float, int]],
         draw: Counter[tuple[float, int]],
         loss: Counter[tuple[float, int]],
-    ) -> RawModelData:
+    ) -> ModelDataDensity:
         """Turns the counts of positions in densities/frequencies
 
         x and y will contain all coordinate values for which data is available.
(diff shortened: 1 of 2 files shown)


vondele:
putting this here as a draft for testing / comments. While conceptually / mathematically more sound, the actual effect on the model is not too large (fortunately I would say).
original approach fitting win rate:
WDL_model_summary_fitDensity

new approach 1: optimizing probability
WDL_model_summary_optimizeProbability

new approach 2: optimizing predicted score
WDL_model_summary_optimizeScore

There might be a bit more difference when --yDataMin 12 or --yPlotMin 12 is reduced, and regions with less data become part of the fit.

scoreWDL.py
```diff
@@ -27,19 +28,24 @@ def save(self, plot_setting: Literal["save+show", "save", "no"]):
...
+    zdraws: list[float]  # draw prob
+    zlosses: list[float]  # loss prob
 
 
 class DataLoader:
     def __init__(self, filenames: list[str]):
         self.filenames = filenames
 
     def load_json(self) -> dict[str, int]:
+        """Load the json, in which the key describes the position (result, move, material),
```
robertnurnberg, line 46: nit, nit ;)

scoreWDL.py
```diff
@@ -107,15 +119,13 @@ def get_raw_model_data(
             zwins.append(win[x, y] / total)
             zdraws.append(draw[x, y] / total)
             zlosses.append(loss[x, y] / total)
-        return RawModelData(xs, ys, zwins, zdraws, zlosses)
+        return ModelDataDensity(xs, ys, zwins, zdraws, zlosses)
 
 
 #
-# fit a model to predict winrate from score and move
+# Model to fit to predict the winrate from score and move
```
robertnurnberg, line 126: did you mean `Model to predict` ?

robertnurnberg:
This looks very good. Running this locally for `updateWDL.json` gives me e.g.
```
> python scoreWDL.py updateWDL.json --plot save --yDataTarget 32 --yDataMin 11 --yDataMax 120 --yPlotMin 18 --modelFitting optimizeScore
Converting scores with NormalizeToPawnValue = 328.
Reading score stats from updateWDL.json.
Retained (W,D,L) = (334275633, 820910739, 341198829) positions.
Fit WDL model based on move.
Initial objective function:  38488672.86008268
/home/rn/git/wdl_model/scoreWDL.py:136: RuntimeWarning: overflow encountered in exp
  return 1.0 / (1.0 + np.exp(-(x - a) / b))
Final objective function:    38488458.97221786
Optimization terminated successfully.
const int NormalizeToPawnValue = 329;
Corresponding spread = 58;
Corresponding normalized spread = 0.17743000244302604;
Draw rate at 0.0 eval at move 32 = 0.992891321288727;
as = ((-1.051 * x / 32 + 9.526) * x / 32 + -15.831) * x / 32 + 336.587
bs = ((-2.645 * x / 32 + 17.442) * x / 32 + -32.639) * x / 32 + 76.258
     constexpr double as[] = {  -1.05135245,    9.52646567,  -15.83068614,  336.58653232};
     constexpr double bs[] = {  -2.64535321,   17.44179845,  -32.63862524,   76.25762993 };
Plotting move/material dependence of model parameters.
Processing done, plotting 2D data.
Saved graphics to WDL_model_summary.png.
```

Note: the overflow error and `const int NormalizeToPawnValue = 329;` compared to `337` with the default fitting in master. Any explanation for this large difference?

As an aside: to help others follow the maths, long term we may want to include the basics in the readme maybe. I have to admit that I am still not 100% sure what `a` and `b` represent, for example, and why `1.0 / (1.0 + np.exp(-(x - a) / b))` is a good ansatz function, so to speak.

vondele:
The overflow error.. is one of the things that needs fixing. It is 'innocent' as the overflow leads to 0 probability, and it is a side effect of the fitting exploring 'unphysical' a,b parameters on the way to the optimum.

The change in NormalizeToPawnValue is to be expected for a different model. It is only a relatively small shifts of the curves, so nothing I'm worried about. It basically is the intrinsic accuracy of this all. It is noteworthy that both probability and score based optimization give very similar values.

concerning the question why this logistic function is the right choice... I definitely is the common choice in this context, there are some explanations out there why this is, .... but admittedly I'm not sure I understand them. 

a, b are basically just setting the mean and variance if you want, a determines the location with 50% winrate and b how quickly things decay around that.

robertnurnberg:
Right, thanks. That already helps. Maybe just for my own benefit I may try to sum up our procedure in a section for the readme... (when I find the time)

vondele:
Rebasing this one, and taking your comments into account. 

I might also give a stab at explaining a bit better in the readme what we're trying to do here.

robertnurnberg:
lgtm

you may want to take out the changed `.pgn` file before you merge

vondele:
so added a section to the readme as well, probably this is now good to go.

robertnurnberg:
Excellent, thanks.

vondele:
so, I would suggest to make one of the new methods the default. Right now, the probability one appears best suited, but I think we need some experience before we decide.

PR #103, opened 2023-11-03
some changes to the new background section

robertnurnberg:
Draft for now, feedback welcome.

Some thoughts:

- We could move the new background section to the bottom of the page, so that the reference to e.g. `NormalizeToPawnValue` makes more sense.
- I have to admit that I am still not 100% clear what precisely `ranging from a simple fit of the observed win rate` means. Is that taking place in the 2D (eval, mom) domain, where the observed win rate in the data at each point, should be well approximated by `win_rate(x,mom)`? So we use some black box that presumably tries to minimize the l^2 error on that discrete 2D grid?
- It would be nice to also explain the six plots we show (another reason to move the section to the end). I think I understand all plots well, and so could come up with a blurb, apart from the bottom left one. What exactly are the dots? The two lines are just the two 1D function `a` and `b`, but what about these discrete points?

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -10,48 +10,6 @@ depend on the engine's evaluation and the move number, and are computed
 from a WDL model that can be generated from fishtest data with the help of
 the scripts in this repository.
 
-## Background
-
-The underlying assumption of WDL model is that the win rate for a position
-can be well modeled as a function of the evaluation of that position.
-The data shows a [logistic function](https://en.wikipedia.org/wiki/Logistic_function) (see also logistic regression)
-gives a good approximation of the win rate as a function of the evaluation.
-
-```
-win_rate(x) = 1 / ( 1 + exp(-(x-a)/b))
-```
-In this equation, the parameters a and b need to be fitted to the data,
-which is the purpose of this repository. a is the evaluation for which a 50% win rate is observed,
-While b indicates how quickly this rate changes with evaluation. A small b indicates that small
-changes in eval quickly turn a 'game on the edge' (i.e. 50% win rate) into a dead draw or a near certain win.
-
-The model furthermore assumes symmetry in evaluation, so that the following quantities follow as well:
-```
-loss_rate(x) = win_rate(-x)
-draw_rate(x) = 1 - win_rate(x) - loss_rate(x)
-```
-
-This information also allows for estimating the game score
-```
-score(x) = 1 * win_rate(x) + 0.5 * draw_rate(x) + 0 * loss_rate(x)
-```
-
-The model is made more accurate by not only taking the evaluation,
-but also the material or game move counter (mom) into account. This dependency is modeled by making the
-parameters a and b a function of mom. The win/draw/loss rates are now 2D functions, while a and b are 1D functions.
-For simplicity the 1D functions a and b are represented as a polynomial of 3rd degree.
-
-The parameters that need to be fit to represent the model completely are thus the 8 parameters that
-determine these two polynomials. For example:
-```
-a(x) = ((-1.719 * x / 32 + 12.448) * x / 32 + -12.855) * x / 32 + 331.883
-b(x) = ((-3.001 * x / 32 + 22.505) * x / 32 + -51.253) * x / 32 + 93.209
-```
-
-Two fit these parameters various approaches exist, ranging from a simple fit of the observed win rate,
-to a somewhat more elaborate maximization of the probability of predicting the correct game outcome
-for the available data.
-
 ## Install
 
 Python 3.9 or higher is required.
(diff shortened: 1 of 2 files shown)


vondele:
The bottom left plot relates to the process how we fit things, at least originally. Originally, we fix the 'mom' (i.e. slice through the 2D data), and compute based on the data the win_rate. This 1D plot is being fit, giving an `a` and `b` for that value for the fixed 'mom'. 
This is 1D fit is exactly what the top left plot shows. Now we repeat that process for each 'mom' value, and that gives us a and b for each mom. These are the dots in the lower left plot. These data points are fitted with the polynomials.

In the new scheme, we still do this 1D fitting curve fitting process, but basically to get a good initial guess for the more advanced fitting in the end (which does a refinement of the a(x) and b(x) based on all data at the same time).

robertnurnberg:
Right, suddenly all seems to make sense. So these 1D slice fits basically give discrete values `a_d(mom)` and `b_d(mom)` for _each_ value of `mom`, and only in a second step we try to fit the degree 3 polynomials through that sea of points?

vondele:
yep

vondele:
and in the new scheme that is just used as an initial guess, and we fit the data globally again...

robertnurnberg:
Right, added more info. I may have become too verbose, just let me know if you think it is too much info.

Also changed some notation to make difference between discrete values `a(mom)` and fitted polynomial `p_a(mom)` clearer.

Btw, out of curiosity. In the top left plot, do we show the `win_rate` function based on `a(32)` and `b(32)`, or based on `p_a(32)` and `p_b(32)`. (Or is it obvious that they must be the same for the anchor?)

vondele:
> Btw, out of curiosity. In the top left plot, do we show the `win_rate` function based on `a(32)` and `b(32)`, or based on `p_a(32)` and `p_b(32)`. (Or is it obvious that they must be the same for the anchor?)

`a(32)` and `b(32)` and it is actually a TODO https://github.com/official-stockfish/WDL_model/blob/ad5aebe92024c52a26a9ee26c8a2d7fbea678a67/scoreWDL.py#L441-L444

Readme.md
```diff
@@ -122,5 +80,93 @@ programs. For example:
...
+score. Once the iteration has converged, the new values for `a(mom)` and 
+`b(mom)` are then the basis for fitting the 1D polynomials `p_a` and `p_b`
+as before.
+
+Finally, note that `x` in the above formulas is the internal engine evaluation
+of a position, which is in general not exposed to the user. By definition
+`x = p_a(mom)` is the internal evaluation with a 50% win rate at material or
+game move counter `mom`. Ideally this `x` should be scaled to the displayed
+evalution `1.0` for every value of `mom`. But for computational simplicity,
+in Stockfish all vaues of `x`, irrespective of the value of `mom`, are scaled
```
vondele, line 151: vaues -> values.

I think this last paragraph can go, it might change if we move further by normalizing different in SF.

vondele:
Overall looks good to me, with possible exception of that one paragraph (but no strong objection either)

Readme.md
```diff
@@ -122,5 +80,93 @@ programs. For example:
...
+the `(x,mom)` data. Denoting these obtained values by `a(mom)` and `b(mom)`, 
+a second step then consists of fitting the 1D polynomials `p_a` and `p_b`
+to these discrete values.
+The options `optimizeProbability` and `optimizeScore` are a bit more
+sophisticated. They take the discrete values `a(mom)` and `b(mom)` provided
+by the above described simple fitting approach as initial guesses for an
+iterative optimization procedure that aims to either
+maximize the probability of predicting the correct game outcome
+for the available data, or to minimize the squared error in the predicted 
+score. Once the iteration has converged, the new values for `a(mom)` and
```
robertnurnberg, line 142: I have some doubts now about that sentence. Is the optimization done on the 1d slices as I claim, or on the polynomials `p_a` and `p_b` directly?

robertnurnberg:
> Overall looks good to me, with possible exception of that one paragraph (but no strong objection either)

I think it helps focus minds and keeps a record of current state of play. Can and will change once SF changes.

Speaking about that, without looking again at your WDLscore branch, could we not simply rescale by `p_a(mom)` for every value of `ply`? Then 1.0 would always be 50% win rate.

Then the WDL repo would only need the polynomial `p_a` to reverse engineer the internal eval?

Readme.md
robertnurnberg: So looking at the code now (while updating the comments there to match readme), I saw that we do both: first local optimization in 1D, but then also a global pass on `p_a` and `p_b`.

robertnurnberg:
Took your comments on board, and also added comments in the python scripts. 

Moved the final paragraph into a separate subsection. We can change that once SF code changes.

PR #105, opened 2023-11-04
minor cleanup and variable renaming

robertnurnberg:
This is a minor cleanup of the recent changes to the python script.

I also took the chance to make the variable names consistent with the readme, in particular now the name `score` is no longer ambiguous. 

Up to rounding errors the changes are non-functional for me.

PS: The rounding errors could be down to the change from `math.log` to `np.log`, but not sure.

### DIFF
--- a/scoreWDL.py
+++ b/scoreWDL.py
@@ -1,12 +1,10 @@
-import json, argparse, numpy as np, matplotlib.pyplot as plt
-from collections import Counter
+import argparse, json, matplotlib.pyplot as plt, numpy as np, time
 from ast import literal_eval
+from collections import Counter
+from dataclasses import dataclass
 from scipy.interpolate import griddata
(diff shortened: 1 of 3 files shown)


vondele:
I think consistency is good, now the script is called scoreWDL ... this ambiguity between eval / score etc is hard to avoid, there is no real consistent usage. But cleanups look good to me. I agree that math.log and np.log can cause the difference (maybe float vs float64)

robertnurnberg:
I think the script name is fine, as in a very wide sense we are "scoring" data. :)

vondele:
can you rebase on current master (some small conflicts)?

robertnurnberg:
Sure, after lunch. :)

robertnurnberg:
Resolved conflict with github GUI, but should be fine.

PR #110, opened 2023-11-05
switch to --modelFitting optimizeProbability

robertnurnberg:
I think the probability based fitting is ready for prod, so to speak.

Attached a plot of the fitted model for my current set of LTC games. Note that I also reduced the lower bounds on move number for fitting and plotting.

PS: The plot seems to indicate that `poly3` may not be the best class of functions for `p_b`, but that is probably a discussion for another day.

updateWDL


vondele:
change of default is good.

I would use the same value for yDataMin and yPlotMin ... better to see what we're using.

I agree that there might be better choices for p_b ... but indeed that's for a later.

robertnurnberg:
Done.
updateWDL

robertnurnberg:
Here with cutoffs at 8.
updateWDL

PR #115, opened 2023-11-13
change internal evals from float to int

robertnurnberg:
This PR might need some discussion before merging.

It changes the current float type for the denormalized internal evals from `float` to `int`. I believe this is more transparent, as these were integers in the engine as well. Moreover, defining them as `int` _may_ open the door to some code optimizations in future, were the 2D data is treating as NumPy 2D arrays, or as SciPy sparse matrices or similar.

The patch is, of course, functional. But only to the extend of the rounding errors from the denormalization. One could maybe argue that the new `int` values are likely closer to the true engine evals than master.

Tested with e.g.
```
> python scoreWDL.py --plot save > master.txt
> diff master.txt patch.txt 
```
to give
```diff
13,14c13,14
< Corresponding normalized spread = 0.17835720039691905;
< Draw rate at 0.0 eval at move 32 = 0.9926807354652202;
---
> Corresponding normalized spread = 0.17856650002022897;
> Draw rate at 0.0 eval at move 32 = 0.9926326548171064;
16,19c16,19
< as = ((1.299 * x / 32 + -5.882) * x / 32 + 16.667) * x / 32 + 323.489
< bs = ((-5.234 * x / 32 + 34.269) * x / 32 + -59.104) * x / 32 + 89.922
<      constexpr double as[] = {   1.29908560,   -5.88159256,   16.66748436,  323.48933024};
<      constexpr double bs[] = {  -5.23429378,   34.26855185,  -59.10404312,   89.92187908 };
---
> as = ((1.297 * x / 32 + -5.871) * x / 32 + 16.652) * x / 32 + 323.096
> bs = ((-5.234 * x / 32 + 34.268) * x / 32 + -59.102) * x / 32 + 89.919
>      constexpr double as[] = {   1.29719238,   -5.87132266,   16.65160741,  323.09634761};
>      constexpr double bs[] = {  -5.23434226,   34.26822898,  -59.10192983,   89.91885990 };
23c23
< Total elapsed time = 20.48s.
---
> Total elapsed time = 20.45s.
```

### DIFF
--- a/scoreWDL.py
+++ b/scoreWDL.py
@@ -35,7 +35,7 @@ class WdlPlot:
 class ModelDataDensity:
     """Count data converted to densities"""
 
-    xs: list[int]  # internal evals in cp
+    xs: list[int]  # internal evals
     ys: list[int]  # moves or material
     zwins: list[float]  # corresponding win probability
     zdraws: list[float]  # draw prob
--- a/scoreWDL.py
+++ b/scoreWDL.py
@@ -106,7 +106,7 @@ class DataLoader:
 
             # convert the cp eval to the internal value
             if NormalizeToPawnValue is not None:
-                eval_internal = eval * NormalizeToPawnValue // 100
+                eval_internal = round(eval * NormalizeToPawnValue / 100)
             else:
                 yDataClamped = min(
                     max(yData, self.NormalizeData["yDataMin"]),
@@ -116,7 +116,7 @@ class DataLoader:
                     yDataClamped / self.NormalizeData["yDataTarget"],
                     *self.NormalizeData["as"],
                 )
-                eval_internal = eval * a // 100
+                eval_internal = round(eval * a / 100)
 
             if result == "W":
                 win[eval_internal, yData] += v
--- a/scoreWDL.py
+++ b/scoreWDL.py
@@ -88,7 +88,7 @@ class DataLoader:
         Counter[tuple[int, int]],
         int,
     ]:
-        """Extract three arrays, win draw and loss, counting positions with a given eval (float) and move/material (int) that are wdl"""
+        """Extract three arrays, win draw and loss, counting positions with a given eval (int) and move/material (int) that are wdl"""
         freq: Counter[tuple[str, int, int, int]] = Counter(
             {literal_eval(k): v for k, v in inputdata.items()}
         )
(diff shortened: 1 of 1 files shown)


robertnurnberg:
I just realized that I wanted to `round` to `int`, rather than doing `//`. Will fix tomorrow and provide new diff.

robertnurnberg:
Fixed. New diff below. As to be expected, the differences are smaller now.
```
13,14c13,14
< Corresponding normalized spread = 0.17835720039691905;
< Draw rate at 0.0 eval at move 32 = 0.9926807354652202;
---
> Corresponding normalized spread = 0.17836150742509121;
> Draw rate at 0.0 eval at move 32 = 0.9926797480716687;
16,19c16,19
< as = ((1.299 * x / 32 + -5.882) * x / 32 + 16.667) * x / 32 + 323.489
< bs = ((-5.234 * x / 32 + 34.269) * x / 32 + -59.104) * x / 32 + 89.922
<      constexpr double as[] = {   1.29908560,   -5.88159256,   16.66748436,  323.48933024};
<      constexpr double bs[] = {  -5.23429378,   34.26855185,  -59.10404312,   89.92187908 };
---
> as = ((1.301 * x / 32 + -5.890) * x / 32 + 16.682) * x / 32 + 323.482
> bs = ((-5.234 * x / 32 + 34.269) * x / 32 + -59.107) * x / 32 + 89.926
>      constexpr double as[] = {   1.30054473,   -5.89004449,   16.68220601,  323.48160643};
>      constexpr double bs[] = {  -5.23424726,   34.26935139,  -59.10734356,   89.92577969 };
23c23
< Total elapsed time = 20.48s.
---
> Total elapsed time = 20.80s.
```

vondele:
Probably that's a good change, but note that we also bin the values when we analyze the pgns, so we don't really recover the original internal evals. We can already remove that binning on analysis, and in particular with the global probability fit, it will remain well behaved even if the binning is set to width 1 (the contour plots will look a little more noisy, but the polynomial fit will remain of good quality).

robertnurnberg:
Right, that is good to know. The more data, the better behaved bin width 1 will be, I guess. Feel free to change the bin width to 1 in the update script. There we could also move the reference commit forward, to better catch the "recent" behaviour of the wdl model.

PR #116, opened 2023-11-15
small edits and variable renamings

robertnurnberg:
Some of the edits would need approval from @vondele and @Disservin . Especially the fact that `win_rate` is now a global function. Personally I think that is more natural, and corresponds to the prominence that function is given in the background section of the readme.

No functional change.

### DIFF
--- a/scoreWDL.py
+++ b/scoreWDL.py
@@ -15,9 +15,9 @@ def win_rate(eval: int | np.ndarray, a, b):
     return stable_logistic((eval - a) / b)
 
 
-def poly3(x: float | np.ndarray, a_3, a_2, a_1, a_0) -> float:
+def poly3(x: float | np.ndarray, c_3, c_2, c_1, c_0) -> float:
     """compute the value of a polynomial of 3rd order in a point x"""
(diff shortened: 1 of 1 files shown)


scoreWDL.py
```diff
@@ -7,9 +7,17 @@
...
+        # returns 1 / (1 + exp(-z)) avoiding possible overflows
+        return np.where(z < 0, np.exp(z) / (1.0 + np.exp(z)), 1.0 / (1.0 + np.exp(-z)))
+
+    return stable_logistic((eval - a) / b)
+
+
+def poly3(x: float | np.ndarray, a_3, a_2, a_1, a_0) -> float:
     """compute the value of a polynomial of 3rd order in a point x"""
-    return ((a * x + b) * x + c) * x + d
+    return ((a_3 * x + a_2) * x + a_1) * x + a_0
```
vondele, line 20: what about calling these coefficients c_3 ... c_0, so we avoid confusion with the a, b that we have in the win_rate.

robertnurnberg: Indeed, that was my original motivation for the renaming here. Will change to c_i as you suggest.

PR #118, opened 2023-11-15
Deal with no data

vondele:


vondele:
useful for exploring WDL models for individual commits, not every commit has data available.

scoreWDL.py
```diff
@@ -800,6 +802,10 @@ def create_plot(self, model_data_density: ModelDataDensity, model: ModelData):
         args.yData,
     )
 
+    if len(win) + len(draw) + len(loss) == 0:
```
robertnurnberg, line 805: later on we skip mom values with `len(ywindata) < 10`, so in theory we could bail out here already if `len(win)<100` or similar. but not sure

robertnurnberg:
how about adding your script to obtain the per-commit raw data of the normalizing constant to the repo?

scoreWDL.py
vondele: I think it is better to deal with that separately. Also have in total just few data points is a bit unlikely, even parsing a single pgn would leads to 10s of data points.

vondele:
I'll make a separate PR, but probably just to share and not to merge at this point, since it is a bit of a hack.

robertnurnberg:
lgtm

PR #119, opened 2023-11-16
WIP: update a range of commits

vondele:
WIP: update a range of commits, one revision at a time

as suggested in https://github.com/official-stockfish/WDL_model/pull/118#issuecomment-1813867965 

Not for merging right now, as this is duplicating quite a bit of the other script we have.

### DIFF
--- a/scoreWDL.py
+++ b/scoreWDL.py
@@ -77,6 +77,8 @@ class DataLoader:
             print(f"Reading eval stats from {filename}.")
             with open(filename) as infile:
                 data = json.load(infile)
+                if not data:
+                    data = {}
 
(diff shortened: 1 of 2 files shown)


robertnurnberg:
So to avoid duplication, and since our update script now allows to pass `--firstrev` and `--lastrev`  from the command line, it might be best to simply call `updateWDL.sh --firstrev $rev --lastrev $rev` over the range of desired commits from a master script, and then rename the produced output files after each run of `updateWDL.sh`.

vondele:
I'll close this, has gotten quite outdated, and is anyway probably not the best way to code this up.

PR #120, opened 2023-11-17
remove final TODOs and refactor code

robertnurnberg:
The refactoring means that we now transparently pass the cli arguments to the functions as needed. Also added some more comments and missing type hints.

I think the plotting and fitting could maybe still be better separated in future, but that is for another PR.

A second pair of eyes from @Disservin would be good.

No functional change (apart from printed text advising what is plotted).

### DIFF
--- a/scoreWDL.py
+++ b/scoreWDL.py
@@ -21,9 +21,19 @@ def poly3(x: float | np.ndarray, c_3, c_2, c_1, c_0) -> float:
 
 
 class WdlPlot:
-    def __init__(self, title: str, pgnName: str):
+    def __init__(
+        self,
+        title: str,
+        pgnName: str,
+        setting: Literal["save+show", "save", "no"],
+        normalize_to_pawn_value: int,
+        yPlotMin: int,
+    ):
         self.title = title
         self.pgnName = pgnName
+        self.setting = setting
+        self.normalize_to_pawn_value = normalize_to_pawn_value
+        self.yPlotMin = yPlotMin
 
         self.fig, self.axs = plt.subplots(
             2, 3, figsize=(11.69 * 1.5, 8.27 * 1.5), constrained_layout=True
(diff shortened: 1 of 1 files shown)


scoreWDL.py
```diff
@@ -81,16 +109,12 @@ def load_json(self) -> Counter[str]:
...
     def extract_wdl(
         self,
-        inputdata: Counter[str],
         moveMin: int,
         moveMax: int,
-        NormalizeToPawnValue: int,
-        NormalizeData: str,
         yDataFormat: Literal["move", "material"],
     ) -> tuple[
         Counter[tuple[int, int]],
```
Disservin, line 120: not visable in the diff but here a return type annotation too much, since you removed it from the returned value

scoreWDL.py
```diff
@@ -354,15 +384,15 @@ def extract_model_data(
         win: Counter[tuple[int, int]],
         draw: Counter[tuple[int, int]],
         loss: Counter[tuple[int, int]],
-        fit,
+        fit: ModelFit,
         plotfunc: Callable[[np.ndarray, list[float], list[float], list[float]], None],
```
Disservin, line 388: not something you changed but it should be `| None` here

Disservin (commented):
Just some general comments which one might act upon in the future.

- The constructors are rather large now which is always a bit annoying. 
- Some stuff is initialized in the constructor to be then just used for one function, a bit unnecessary.
- Would be nice if the code goes into a more stateless direction, cant think of a better term right but avoiding computed data to be class members

Again some rather general stuff which is not strictly related to this pr, but I thought i might just share my thoughts. 

Btw do you use a type checker ? That can spot some issues early on, there are still a few ones in the code right now. I just have mypy running in my IDE and that's it.

robertnurnberg:
Thanks for your comments. I had run `mypy` once on master a few commits back and was deterred by the many errors reported. I have now run it on the PR and fixed some things related to it.

I agree that from an OOP PoV the script still has some way to improve, but I think we are slowly getting there.

> Some stuff is initialized in the constructor to be then just used for one function, a bit unnecessary.

I usually try to answer the question: does the "stuff" form part of the object, and so is an attribute of it, or is it something external. In the former case I make it member data, even if it is then only used inside one function. In the latter case I would pass it as a parameter to the function.

PS: One `mypy` error I still get is `scoreWDL.py:817: error: Argument 3 to "DataLoader" has incompatible type "Optional[Any]"; expected "str"`. I have ignored it for now. Is there an elegant way to silence it?

Disservin:
> PS: One `mypy` error I still get is `scoreWDL.py:817: error: Argument 3 to "DataLoader" has incompatible type "Optional[Any]"; expected "str"`. I have ignored it for now. Is there an elegant way to silence it?

just fix the type properly, no? :D
```python
class DataLoader:
    def __init__(
        self, filenames: list[str], NormalizeToPawnValue: int | None, NormalizeData: str | None
    ):
```

robertnurnberg:
Perfect, thanks. :)

PR #121, opened 2023-11-17
move log files into separate directory

Disservin:


robertnurnberg:
Note that in master the location gives the directory in which command was executed. May or may not make debugging easier.

Also, some log files in master had the same name but different directories, they would now overwrite each other.

For me the way it is in master is fine, but I have no strong objections against the PR.

vondele:
would it make sense to have all output (including the json and the png) of the script in a subdirectory? Would avoid the issue of overwriting the .png that is part of the repo, and keep output together, so that it is clear which image and which logs belong together.

robertnurnberg:
Good idea. Note that png is no longer an issue, as we changed default name a while back. :)

Disservin:
Sounds like a good idea vondele and yeah just realized that the pull logs overwrite each other. Still having the logs in the same place would be nice

Disservin:
Mh okay so not sure what to think :D With the last commit it should put all files into `update/` but it adds quite some overhead.  
To get back to one of the previous points
> Note that in master the location gives the directory in which command was executed. May or may not make debugging easier.

I dont think it's even particular easy right now... I find it quite weird that all output is separated into different log files, if you want to debug you need to know in which order the files are created and then go through them sequentially, while there might be still old log files in the directory. 
I'd just output everything to stdout and let the user redirect the script output into a log file.

robertnurnberg:
Not saying what is better or worse, just for historical perspective: the way update works is based on similar scripts in the matetrack repo.

For debugging in master one can see where output stops in the update log, then check other logs until that point. Time stamps of log files also help.

Disservin:
So what do we think ? I find the `$outpath` juggling quite annoying. I'd rather just keep the separate directory structure and do
`exec &> update/update.log`.

vondele:
so having a single log for the output is not so convenient, like I don't need the output of git clone to the terminal, I only care about that when it fails. The output of the wdl fit is almost always interesting, and that's good when it is available next to the figures. I agree currently the script looks a little overloaded. For one, I'm uncertain why we need `"$logpath"` ... is that to deal with directories with spaces?

robertnurnberg:
We could simply put all in the output path, including logs. For those with same names, we can add e.g. the repo name to the log file name.

vondele:
I think all output in one path makes sense to me.

Disservin:
I think you imagined it like this ?

robertnurnberg:
Not sure if the best place for the cloned repos is in `update`, one could imagine a version of this script being done with different setting, and output being put in different output directory. Think the WIP PR, for example.

Small nit: why sometimes `./` before the dir name, and sometimes not? Finally: do not forget to run `make format` at the end.

Disservin:
Ah thanks ;)

> Not sure if the best place for the cloned repos is in update, one could imagine a version of this script being done with different setting, and output being put in different output directory. Think the WIP PR, for example.

Well I think in such a scenario one can easily just make the outpath a cli parameter to create a distinct output directory.

robertnurnberg:
Agreed. But it would result in new clones of the repos for each different output dir?

robertnurnberg:
So what I am trying to say is that it may be best to leave the two repo directories where they are. I would also store the `fixfen` EPD files in the `books` repo dir. This way there are available for various different update runs.

updateWDL.sh
```diff
@@ -48,12 +52,13 @@ fixfen="fixfen_$bookhash.epd"
 if [[ ! -e "$fixfen.gz" ]]; then
```
robertnurnberg, line 52: missing new dir

Disservin:
okay @robertnurnberg what do you say we do here? I think ultimately I'd simply like to avoid cluttering the entire root directory with all related update files.

robertnurnberg:
Hm, I don't really mind the status quo. The update files are `updateWDL.json`, `updateWDL.png` and a few log files?

I am planning a few more changes to the update script to help prepare for when https://github.com/official-stockfish/Stockfish/pull/4920 gets merged. So maybe we can wait on this PR for a bit longer until the transition is done, and then get back to it?

Disservin:
sure thing ill close it in the meantime and then we can see some other time

PR #123, opened 2023-11-18
simplification: change wdl tuple to probabilities

robertnurnberg:
I guess the original function was used (also) in different contexts? @vondele  But now we can just treat wdl as probabilites. Surprisingly, the text output for my tests remains unchanged, even though the slightly functional nature of the PR means the png file changes slightly.


vondele:
looks good, some conflicts after the latest merge

robertnurnberg:
Resolved (and thanks for merging already). I'd have thought that git would be clever enough to resolve/avoid this simple conflict. ;)

PR #124, opened 2023-11-18
rename popt_as to coeffs_a

robertnurnberg:
This brings the code more in line with the readme, and makes it easier to distinguish between p_a and a.

I also updated several comments and would want to have confirmation from @vondele that these still make sense, before the PR being merged.

No functional change.


vondele:
yes, looks good

PR #126, opened 2023-11-18
deal with unphysical parameters centrally

robertnurnberg:
I think the model optimizations will avoid a<=0 naturally, so this PR only deals with b <= 0. Here as an extra guard we also check for b < 10^{-8}, and if so set the win rate to 0 or 1, depending on the sign of the argument.

I cannot test this for "degenerate" data, which may have necessitated the original safeguard in the first place. So feedback/tests by @vondele would be appreciated. In my tests the PR is nonfunctional.

### DIFF
--- a/scoreWDL.py
+++ b/scoreWDL.py
@@ -12,7 +12,8 @@ def win_rate(eval: int | np.ndarray, a, b):
         # returns 1 / (1 + exp(-z)) avoiding possible overflows
         return np.where(z < 0, np.exp(z) / (1.0 + np.exp(z)), 1.0 / (1.0 + np.exp(-z)))
 
-    # guard against unphysical values
+    # guard against unphysical values, treating small and negative values as 0
+    # during optimizations this will guide b (back) towards positive values
     if b < 1e-8:
         return np.where(eval - a < 0, 0, 1)
 
--- a/scoreWDL.py
+++ b/scoreWDL.py
@@ -12,6 +12,10 @@ def win_rate(eval: int | np.ndarray, a, b):
         # returns 1 / (1 + exp(-z)) avoiding possible overflows
         return np.where(z < 0, np.exp(z) / (1.0 + np.exp(z)), 1.0 / (1.0 + np.exp(-z)))
 
+    # guard against unphysical values
+    if b < 1e-8:
+        return np.where(eval - a < 0, 0, 1)
+
     return stable_logistic((eval - a) / b)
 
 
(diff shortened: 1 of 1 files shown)


robertnurnberg:
Thinking a bit more about this, we could put in a final check after p_a and p_b are found, to see if they are both positive on the whole data range? And report an error if so?

vondele:
not sure this is fully correct as a replacement. Actually with the stable_logistic, it is probably enough to just guard against abs(b) < 1e-8, but the result depends on sign(eval - a) * sign(b).

robertnurnberg:
Yeah, I wanted to "forbid" negative values of b, so they are treated as 0. This should naturally guide optimization towards positive values?

vondele:
OK, I see, yeah, that might work with a change to the comment.

robertnurnberg:
Added a comment.

vondele:
hmm, doesn't seem to work for me. I'm getting a warning and an error:
```
/home/vondele/chess/vondele/WDL_model/scoreWDL.py:13: RuntimeWarning: overflow encountered in exp
  return np.where(z < 0, np.exp(z) / (1.0 + np.exp(z)), 1.0 / (1.0 + np.exp(-z)))
/home/vondele/chess/vondele/WDL_model/scoreWDL.py:13: RuntimeWarning: invalid value encountered in double_scalars
  return np.where(z < 0, np.exp(z) / (1.0 + np.exp(z)), 1.0 / (1.0 + np.exp(-z)))
 ```
I think that's related to the explanation here https://stackoverflow.com/questions/52622172/numpy-where-function-can-not-avoid-evaluate-sqrtnegative both arguments are evaluated in the where, even though the right one is picked in the end.

The other one is:
```
  File "/home/vondele/chess/vondele/WDL_model/scoreWDL.py", line 808, in <module>
    wdl_model.create_plot(model_data_density, model)
  File "/home/vondele/chess/vondele/WDL_model/scoreWDL.py", line 634, in create_plot
    zz = model_wdl_tuple(
  File "/home/vondele/chess/vondele/WDL_model/scoreWDL.py", line 39, in model_wdl_tuple
    w = win_rate(eval, a, b)
  File "/home/vondele/chess/vondele/WDL_model/scoreWDL.py", line 17, in win_rate
    if b < 1e-8:
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```
(note this is tested with https://github.com/official-stockfish/WDL_model/pull/128 merged).

robertnurnberg:
Thanks for testing and reporting. I had the warning in development but thought it was due to my sparse dataset. Will investigate.

Will also look at the other error later today. Can you give me your command line that triggers it?

robertnurnberg:
Oh, I think I missed the second error because I had not merged this patch into my branch for the other pr. Sorry about that.

robertnurnberg:
Opened https://github.com/official-stockfish/WDL_model/pull/129 as possible fixes.

PR #128, opened 2023-11-18
simplify contour plotting

robertnurnberg:
No functional change.

### DIFF
--- a/scoreWDL.py
+++ b/scoreWDL.py
@@ -591,9 +591,6 @@ class WdlModel:
             self.plot.axs[1, 0].set_title("Winrate model parameters")
             self.plot.axs[1, 0].set_ylim(bottom=0.0)
 
-        # now generate contour plots
-        contourlines = [0, 0.05, 0.1, 0.25, 0.5, 0.75, 0.9, 0.97, 1.0]
-
         ylabelStr = self.yData + " (1,3,3,5,9)" * bool(self.yData == "material")
         for i in [0, 1] if self.modelFitting != "None" else [0]:
             for j in [1, 2]:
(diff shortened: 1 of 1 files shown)


scoreWDL.py
```diff
@@ -604,92 +601,53 @@ def create_plot(self, model_data_density: ModelDataDensity, model: ModelData):
...
+            xmax = (((3 - j) * self.plot.normalize_to_pawn_value) // 100 + 1) * 100
+            ymin, ymax = self.plot.yPlotMin, self.yDataMax
+            grid_x, grid_y = np.mgrid[xmin:xmax:30j, ymin:ymax:22j]  # use a 30x22 grid
+
+            # data
+            zz = model_data_density.zdraws if j else model_data_density.zwins
+            zz = griddata(points, zz, (grid_x, grid_y), method="linear")
+            cp = self.plot.axs[0, 1 + j].contourf(grid_x, grid_y, zz, contourlines)
+            if j == 0:
+                self.plot.fig.colorbar(cp, ax=self.plot.axs[:, -1], shrink=0.618)
```
robertnurnberg, line 621: @vondele 
I noticed that when moving this `colorbar` line to _after_ the `for` loop, the png changes. So I assume the colour palette used in plots (0,1) and (2,2) is slightly different? I have no other explanation.
Anyway, if we do not mind the difference, then moving this line to below the `for` loop (without an `if`) would simplify the code slightly.

robertnurnberg:
So, once https://github.com/official-stockfish/WDL_model/pull/129 is merged, this should be nonfunctional. Regarding the colorbar, I would say to merge this PR as is, and deal with that issue later on.

Of course, I plan to do further simplifications of the plotting part: doing all four contour plots in one loop, and moving the plotting functionality into the `WdlPlot` class.

robertnurnberg:
In fact, I am going to close this one and open follow-up with a more comprehensive simplification.

PR #136, opened 2023-11-21
use round not trunc for NormalizeToPawnValue

robertnurnberg:
I believe rounding is the more correct approach. In the extreme case, in master a value like `328.9` is truncated to `328`.

**Important note**: Requires a corresponding fix to the `static_assert` in stockfish code with the next WDL update, and similarly for other engines that use this repo for their internal WDL model and have the same assert as stockfish.

Example:
```
> cat master.txt
Converting evals with NormalizeToPawnValue = 328.
Reading eval stats from scoreWDLstat.json.
Retained (W,D,L) = (6872929, 16756625, 6971320) positions.
Fit WDL model based on move.
Warning: Too little data for move value 3, skip fitting.
Warning: Too little data for move value 4, skip fitting.
Warning: Too little data for move value 5, skip fitting.
Warning: Too little data for move value 6, skip fitting.
Warning: Too little data for move value 7, skip fitting.
Warning: Too little data for move value 8, skip fitting.
const int NormalizeToPawnValue = 335;
Corresponding spread = 59;
Corresponding normalized spread = 0.17836150742509121;
Draw rate at 0.0 eval at move 32 = 0.9926797480716687;
as = ((1.301 * x / 32 + -5.890) * x / 32 + 16.682) * x / 32 + 323.482
bs = ((-5.234 * x / 32 + 34.269) * x / 32 + -59.107) * x / 32 + 89.926
     constexpr double as[] = {   1.30054473,   -5.89004449,   16.68220601,  323.48160643};
     constexpr double bs[] = {  -5.23424726,   34.26935139,  -59.10734356,   89.92577969 };
Preparing contour plots and plots of model parameters.
Saved graphics to scoreWDL.png.
```

`diff master.txt patch.txt`:
```diff
11,12c11,12
< const int NormalizeToPawnValue = 335;
< Corresponding spread = 59;
---
> const int NormalizeToPawnValue = 336;
> Corresponding spread = 60;
```

### DIFF
--- a/scoreWDL.py
+++ b/scoreWDL.py
@@ -48,7 +48,7 @@ class WdlData:
         if self.NormalizeData is not None:
             self.NormalizeData = json.loads(self.NormalizeData)
             self.NormalizeData["as"] = [float(x) for x in self.NormalizeData["as"]]
-            self.normalize_to_pawn_value = int(sum(self.NormalizeData["as"]))
+            self.normalize_to_pawn_value = int(sum(self.NormalizeData["as"]) + 0.5)
         else:
             self.normalize_to_pawn_value = args.NormalizeToPawnValue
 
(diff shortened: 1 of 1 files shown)


vondele:
so, I would park this one for now, and start doing this at the point where we update the WDL model in SF, introduce the ply clamp and this different rounding.

robertnurnberg:
Happy with whatever you prefer....

....but: wouldn't the natural order be to do it first here, then create new WDL model with that output, and then adapt sf?

vondele:
sure, first here, but essentially at the same.... so that people using this repo and SF code are not confused about why it is not working.

robertnurnberg:
(Unless you want to leave it right before reading off the new WDL model, to not forget the change in SF.)

Btw, I think signs are that next WDL model in SF will use dynamic normalization (if https://github.com/official-stockfish/fishtest/pull/1848 gets merged by then), and then we could drop NormalizeToPawnValue altogether in SF. Dropping it would remove some lines of unneeded code, but would make tracking the inflation/deflation of internal eval a bit harder. But we could do that here within this repo, maybe together with your commit-wise tracking of the best fitted WDL model. So one curve `p_a(32)` hard-coded in SF over time, the other curve `p_a(32)` from WDL model fitted from fishtest data.

Gotta run ... :)

robertnurnberg:
Just for the record, I used this branch for https://github.com/official-stockfish/Stockfish/pull/5002 , to get the rounded value for `NormalizeToPawnValue`.

Disservin:
@robertnurnberg once i merge the stockfish pr (any time soon), i'll undraft this and merge it too.

PR #140, opened 2023-11-26
Normalize objective functions

vondele:
go from sum of squared error to root mean squared error in the score, divide log prob by the number of positions (i.e. use geometric mean)

### DIFF
--- a/scoreWDL.py
+++ b/scoreWDL.py
@@ -1,4 +1,4 @@
-import argparse, json, matplotlib.pyplot as plt, numpy as np, time
+import argparse, json, matplotlib.pyplot as plt, numpy as np, time, math
 from ast import literal_eval
 from collections import Counter
 from dataclasses import dataclass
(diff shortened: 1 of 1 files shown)


vondele:
should be non-functional apart from numerics.

robertnurnberg:
Instead of manually added to counter, could just read of the lengths of the data containers?

Btw: I am working on a major rewrite of the internal data structures, going full numpy. Hope is for cleaner and clearer code that will also be significantly faster. Stay tuned...

scoreWDL.py
```diff
@@ -1,4 +1,4 @@
-import argparse, json, matplotlib.pyplot as plt, numpy as np, time
+import argparse, json, matplotlib.pyplot as plt, numpy as np, time, math
```
robertnurnberg, line 1: Using `np.sqrt` is also a possibility. ;)

robertnurnberg:
Of course, I agree with the idea behind the pr.

PR #142, opened 2023-12-05
output coefficients in clang format

robertnurnberg:
Master has a rogue white space in the printed list of `bs` coefficients. PR removes that and also simplifies the code.

No functional change.


Disservin:
why draft?

robertnurnberg:
I worked on a small improvement. ;)

Disservin:
tbf i found the commit before a bit more readable also i guess if we already change it the output should match stockfishs clang format

robertnurnberg:
so improvement was avoiding code repetition (here the code formatter)

the latter has now been changed to match clang format

PR #141, opened 2023-11-30
go full numpy

robertnurnberg:
This PR is a nearly complete rewrite of the python script. In fact, for the maintainers it might be easier to look at the new `scoreWDL.py` than at the diff.

The main change is to the underlying data structure. The win/draw/loss counts (and densities) are now collected in 2D numpy arrays, with "coordinates" `(mom, eval)`. Of course, the indices to numpy arrays must be nonnegative, and so internally we need to work with offsets to go from `(mom, eval)` to `(mom_idx, eval_idx)`.

The restructuring allowed for a significant simplification of the whole code. The different aspects of the script are now also clearly separated:

- all the raw data from the games are stored in the class `WdlData`
- all the (fitted) model parameters are stored in the class `WdlModel`
- all the plotting functionality is provided by `WdlPlot`, and the plotting is only invoked after all the fitting is done

The PR is nearly nonfunctional, with the following exceptions:

- rounding errors lead to slightly different model parameters
- two labels for plotting and text output changed from "as = ", "bs = " to "p_a = ", "p_b = "
- for the cli option `--NormalizeData` the eval cut-off changed from 400 in cp to an appropriately scaled cut-off for the internal eval (which is nonfunctional for a static renormalization)

The code still contains several `TODO`. Some of these are trivial and can be done soon after merging, others require more work and time.

Possible improvements/future work:

- look into using sparse matrices as the underlying data structure (in place of full 2D numpy arrays)
- there are now three places, where parallelization with e.g. `concurrent.futures` could yield a speed-up: loading the json, fitting a and b locally, computing the objective functions in the two optimization stages
- maybe some more type hint love from @Disservin  ;)   (for now, all nontrivial function definitions have type hints for their arguments)

PR #144, opened 2023-12-06
bug fix

robertnurnberg:
This fixes the bug discussed in https://github.com/official-stockfish/WDL_model/pull/141#issuecomment-1842440587

### DIFF
--- a/scoreWDL.py
+++ b/scoreWDL.py
@@ -43,6 +43,8 @@ class WdlData:
 
     def __init__(self, args):
         self.yData = args.yData
+        self.yDataMin = args.yDataMin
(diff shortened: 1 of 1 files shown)


robertnurnberg:
Note that this changes the behaviour for the global optimization routines compared to https://github.com/official-stockfish/WDL_model/commit/ece3fcef9fc7dea9cf1d1ad493b14908a331e534 I do not think the global optimization over _all_ game data within the `move` and `eval` bounds was intended behaviour, as it ignored the limits given by `yDataMin` and `yDataMax`. 

In addition the `(W,D,L)` counts change slightly, as compare to master some game data are rejected at the data load stage.

PR #145, opened 2023-12-07
unify mom min/max cli parameters

robertnurnberg:
In new master, when `--yData` is `"move"`, then `--moveMin` and `--yDataMin` have the same meaning, and similarly for `Max`. This PR unifies this, avoiding possible misunderstandings.

PR #146, opened 2023-12-10
rename yData to mom, plus other changes

robertnurnberg:
The name `yData` served its purpose in the transition from the two original scripts (one for move, one for material) to the code base we have now. IMO we could now transfer to `mom`, as per code comments and readme.

This PR:
-  ~~is based on~~ includes the changes in https://github.com/official-stockfish/WDL_model/pull/145, ~~to make it easier to rebase once that is merged (alternatively maintainers could merge this PR directly, and I will close https://github.com/official-stockfish/WDL_model/pull/145)~~
- performs the mentioned renaming from `yData` to `mom` (e.g. `--yData` becomes `--momType`)
- deals with one of the `TODO`, by introducing `--winMin`
- allows different input/output WDL models for the dynamic rescaling: i.e. one can be based on move, the other can be based on material (to this end, `NormalizeData` also gets the entry `momType`)
- provides more help for `--help` and changes the text for some of the options accordingly

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -76,9 +76,9 @@ programs. For example:
 
 - `scoreWDLstat --matchEngine <regex>` : extracts WDL data only from the
    engine matching the regex
-- `python scoreWDL.py --yDataTarget 30` : chooses move 30 (rather than 32)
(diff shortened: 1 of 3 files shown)


Disservin (approved):
lgtm

PR #151, opened 2023-12-18
set default --materialMax to 78

robertnurnberg:
This is a left-over change from https://github.com/official-stockfish/WDL_model/pull/146, and sets the default value for `--materialMax` to the material count of the starting position.

As this could potentially also affect the fitting by move (due to the filter applied to the json data), merging this PR should await approval by @vondele. (Basically in master hardly any position will be filtered based on material count, whereas this PR may filter some games with very early promotions, but I think that is highly unlikely.)

### DIFF
--- a/scoreWDL.py
+++ b/scoreWDL.py
@@ -589,7 +589,7 @@ if __name__ == "__main__":
     parser.add_argument(
         "--materialMax",
         type=int,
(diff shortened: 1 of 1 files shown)


vondele:
I think this is fine, the number of games filtered will be very small.

PR #153, opened 2023-12-21
enable dynamic renormalization in updateWDL.sh

robertnurnberg:
This PR prepares the update script for the future, when https://github.com/official-stockfish/Stockfish/pull/4920 will be merged into SF.

Some comments:
- The PR relies on `NormalizeToPawnValue` *not* being present in `src/uci.h` for SF revisions that do the dynamic eval rescaling. Edit: It now relies on `const in NormalizeToPawnValue` not being present in `src/uci.h` and `constexpr int  NormalizeToPawnValue` not being present in `src/uci.cpp`.
- For those revisions, the corresponding value of `NormalizeToPawnValue` is computed on the fly, by summing the coefficients in `as`. The final sum is *rounded*, so that after merging https://github.com/official-stockfish/WDL_model/pull/136 this will be consistent with the python script. Note also that for un-doing the dynamic rescaling, the value `NormalizeToPawnValue` serves only a cosmetic purpose now. It is used at the end of the update script to indicate to the user if that value has changed from SF to the new model, or not.
- If this new update script is called with the same sha for `--firstrev` and `--lastrev` it allows for extracting the newly fitted `NormalizeToPawnValue` for both static and dynamic renormalization revisions of SF. This may be useful when adapting https://github.com/official-stockfish/WDL_model/pull/119

For as long as https://github.com/official-stockfish/Stockfish/pull/4920 is not merged into SF, this PR is nonfunctional. Once that PR is merged, the update script needs to be pointed to that new commit in `default_firstrev`, and it should then seamlessly work for the dynamic eval rescalings.

### DIFF
--- a/scoreWDL.py
+++ b/scoreWDL.py
@@ -53,6 +53,10 @@ class WdlData:
             self.normalize_to_pawn_value = int(sum(self.NormalizeData["as"]))
             if not "momType" in self.NormalizeData:
                 self.NormalizeData["momType"] = "move"
+            assert self.NormalizeData["momType"] in [
+                "move",
+                "material",
+            ], "Error: momType must be move or material."
         else:
             self.normalize_to_pawn_value = args.NormalizeToPawnValue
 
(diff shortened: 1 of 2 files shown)


robertnurnberg:
Latest commit to this PR fixes https://github.com/official-stockfish/WDL_model/issues/156.

Disservin:
Can we have the grep be completely source file independent? I.e moving the constant around would break the current code?

robertnurnberg:
I think to avoid false positives, the current way is best.

We can make it more general the next time the constant is moved around. (I am running the update script every day, so would spot when things go wrong next time.)

Soon-ish the constant will disappear from SF code, once we switch to dynamic rescaling.

PR #154, opened 2024-01-11
scoreWDLstat.cpp cleanup

Disservin:
reorder some local variables and abstract cli and filtering, maybe a bit overengineered @robertnurnberg what do you say? Would you also be so kind as to test if i screwed up anywhere?


robertnurnberg:
Hm, looks like this breaks something. Running the update script locally reports `Found 0 .pgn(.gz) files, creating 0 chunks for processing.`, while master works fine.

Disservin:
Oh is this because of the refactoring?

vondele:
looks like this one needs a rebase.

PR #172, opened 2024-03-17
prepare updateWDL.sh for material based WDL model

robertnurnberg:
In preparation for https://github.com/official-stockfish/Stockfish/pull/5121 (so to be merged after this is merged into SF).

Also remove a no longer used book (for SPRT) from the book match pattern.


Disservin:
btw can you update the readme parts about `NormalizeToPawnValue` @robertnurnberg ?

robertnurnberg:
Yes. Was planning to do that in a separate pr. Once SF uses the new normalization.

PR #175, opened 2024-03-20
update chess-library

Disservin:
Should be non functional


robertnurnberg:
Nonfunctional for me when run with `./updateWDL.sh --firstrev e67cc979fd2c0e66dfc2b2f2daa0117458cfc462`

PR #176, opened 2024-03-20
reflect WDL model update in SF

robertnurnberg:
This only deals with the scripts. ~~and the figure.~~

Will update the text of the readme in a separate PR.

### DIFF
--- a/scoreWDL.py
+++ b/scoreWDL.py
@@ -604,7 +604,7 @@ if __name__ == "__main__":
     parser.add_argument(
         "--moveMin",
         type=int,
-        default=3,
+        default=1,
         help="Lower move number limit for filter applied to json data.",
     )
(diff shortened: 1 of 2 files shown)


robertnurnberg:
I have now also changed all the default parameters to reflect what is needed for the _current_ WDL model of Stockfish.

This may mean that developers of other engines need to specifically state their desired parameters, if they relied on the default values so far.

Probably needs a second pair of eyes and approval from @vondele

Disservin:
mh personally i'd just keep the current defaults

robertnurnberg:
That is an option. But note the "clunky" options needed the transform the cp evals from the json data correctly. Ideally we would not want to have to type that in every time we deal with the new SF-evals.

PR #177, opened 2024-03-20
reflect material based fitting in the readme

robertnurnberg:
This PR updates the text of the readme to reflect https://github.com/official-stockfish/Stockfish/pull/5121.


vondele:
I think having the default right for SF while documenting the former ply based formula as an example in the readme (as is done with the two pending PRs) is probably best.

PR #179, opened 2024-03-23
show 0MB download size

robertnurnberg:
This PR shows `0MB` for tar files smaller than 1MB. 

Master:
```
Downloading .pgns.tar file with 120 games to ./pgns/24-03-21/65fca00c0ec64f0526c5189e/ ...
```

Patch:
```
Downloading 0MB .pgns.tar file with 120 games to ./pgns/24-03-21/65fca00c0ec64f0526c5189e/ ...
```

### DIFF
--- a/download_fishtest_pgns.py
+++ b/download_fishtest_pgns.py
@@ -148,8 +148,8 @@ while True:
         url = "https://tests.stockfishchess.org/api/run_pgns/" + test + ".pgns.tar"
         try:
             response = urllib.request.urlopen(url)
-            mb = int(response.getheader("Content-Length", 0)) // (2**20)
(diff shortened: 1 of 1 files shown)


vondele:
while it is a detail, is it possible to round up? If there are 120 games, I'd expect 1MB output. No big deal if that's hard.

robertnurnberg:
With latest commit this becomes `Downloading 123KB .pgns.tar file with 120 games to ./pgns/24-03-21/65fca00c0ec64f0526c5189e/ ...` for patch.  :)

(master had always rounded down to full MB, this patch now also switches to decimal system, as it is used in general for memory sizes on hard disks etc)

vondele:
Hey, it fails on my Exabyte sized pgn collections ;-).

we'll deal with it when it happens...

PR #180, opened 2024-03-24
in plots of p_a and p_b also show old p_a, if available

robertnurnberg:
This is now possible, if old and new model used the same `momType`.

I made new `p_a` bold, and old `p_a` dashed with a lighter red. Can easily be changed, ofc.

### DIFF
--- a/scoreWDL.py
+++ b/scoreWDL.py
@@ -498,14 +498,29 @@ class WdlPlot:
             self.axs[1, 0].plot(
                 model.ms,
                 poly3(model.ms / model.momTarget, *model.coeffs_a),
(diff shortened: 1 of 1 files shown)


robertnurnberg:
I used this to compare old and new, for the (small) data set of games since last WDL model change. Because of the very significant reduction in the eval rescaling constant, I wanted to see if I have made a mistake somewhere. I cannot find a bug in the code, so the dramatic change is probably down to some recent eval changing commits to SF?

scoreWDL

vondele:
is there some better label than 'old' ? Is this 'Model used for data generation' ?

PR #185, opened 2024-04-26
adapt download script to new fishtest API

robertnurnberg:
This ensures the download script works with after the latest fishtest update, see https://github.com/official-stockfish/fishtest/pull/1955

I prepared this quickly, so will need to get back to obtaining the size of the download correctly at a later stage.

### DIFF
--- a/download_fishtest_pgns.py
+++ b/download_fishtest_pgns.py
@@ -1,4 +1,4 @@
-import argparse, datetime, json, os, re, tarfile, urllib.request
+import argparse, datetime, json, os, re, urllib.request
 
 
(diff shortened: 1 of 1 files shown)


vondele:
looks good to me, it might be no longer possible to obtain the size though?

robertnurnberg:
> looks good to me, it might be no longer possible to obtain the size though?

Possibly. I have asked Pasquale if it can be broadcast from the server. If not, we can remove that code in the downloader.

Disservin:
ah that's why rob asked about the `Content-Length`..

PR #191, opened 2024-05-22
increase default materialMin to 17

robertnurnberg:
As per the plots and discussion on [discord](https://discord.com/channels/435943710472011776/900771437177094156/1242729158057005158), it was decided it is probably best to increase the lower limit for our material count data from 10 to 17.

At low material count values the data is likely too sparse to create meaningful fits. 

Plots of master vs patch below.
master

patch

### DIFF
--- a/scoreWDL.py
+++ b/scoreWDL.py
@@ -637,7 +637,7 @@ if __name__ == "__main__":
     parser.add_argument(
         "--materialMin",
         type=int,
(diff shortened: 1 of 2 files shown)


robertnurnberg:
Note: I is probably a good idea to also update the model in SF in the near future, so that we do not forget to change the lower value of the clamp also there. I can take care of that.

PR #193, opened 2024-05-26
allow filtering based on estimated nElo difference

robertnurnberg:
This PR allows filtering the pgn data based on estimated nElo difference, as discussed on discord with @vondele.

I took the computation of the nElo difference from [fast-chess](https://github.com/Disservin/fast-chess/blob/4dbbc9f58bb4fd7498d92b0a27ef385f2a52f9bc/src/elo/elo_pentanomial.cpp#L25).

For the update script I set the default maximal nElo difference to 5.

Example usage:
```
> ./scoreWDLstat -r --EloDiffMax 0 --matchRev 61acbfc7d310ed6044ba4fc5ef91a6c382d1c9a6 --matchBook UHO_Lichess_4852_v1.epd
Looking (recursively) for pgn files in ./pgns
Found 164289 .pgn(.gz) files in total.
Filtering pgn files matching the book name UHO_Lichess_4852_v1.epd
Filtering pgn files matching revision SHA 61acbfc7d310ed6044ba4fc5ef91a6c382d1c9a6
Filtering pgn files with nElo in [-0, 0]
Found 0 .pgn(.gz) files, creating 0 chunks for processing.
```
```
> ./scoreWDLstat -r --EloDiffMax 1.95 --matchRev 61acbfc7d310ed6044ba4fc5ef91a6c382d1c9a6 --matchBook UHO_Lichess_4852_v1.epd
Looking (recursively) for pgn files in ./pgns
Found 164289 .pgn(.gz) files in total.
Filtering pgn files matching the book name UHO_Lichess_4852_v1.epd
Filtering pgn files matching revision SHA 61acbfc7d310ed6044ba4fc5ef91a6c382d1c9a6
Filtering pgn files with nElo in [-1.95, 1.95]
Found 9 .pgn(.gz) files, creating 9 chunks for processing.
```
```
> ./scoreWDLstat -r --EloDiffMax 100 --matchRev 61acbfc7d310ed6044ba4fc5ef91a6c382d1c9a6 --matchBook UHO_Lichess_4852_v1.epd
Looking (recursively) for pgn files in ./pgns
Found 164289 .pgn(.gz) files in total.
Filtering pgn files matching the book name UHO_Lichess_4852_v1.epd
Filtering pgn files matching revision SHA 61acbfc7d310ed6044ba4fc5ef91a6c382d1c9a6
Filtering pgn files with nElo in [-100, 100]
Found 15 .pgn(.gz) files, creating 15 chunks for processing.
```
```
> ./scoreWDLstat -r --EloDiffMin -10 --EloDiffMax 1.95 --SPRTonly --matchRev 61acbfc7d310ed6044ba4fc5ef91a6c382d1c9a6 --matchBook UHO_Lichess_4852_v1.epd
Looking (recursively) for pgn files in ./pgns
Found 164289 .pgn(.gz) files in total.
Filtering pgn files matching the book name UHO_Lichess_4852_v1.epd
Filtering pgn files matching revision SHA 61acbfc7d310ed6044ba4fc5ef91a6c382d1c9a6
Filtering pgn files with nElo in [-10, 1.95]
Found 15 .pgn(.gz) files, creating 15 chunks for processing.
```

Without the filtering option behaviour is like master:
```
> ./scoreWDLstat -r --matchRev 61acbfc7d310ed6044ba4fc5ef91a6c382d1c9a6 --matchBook UHO_Lichess_4852_v1.epd
Looking (recursively) for pgn files in ./pgns
Found 164289 .pgn(.gz) files in total.
Filtering pgn files matching the book name UHO_Lichess_4852_v1.epd
Filtering pgn files matching revision SHA 61acbfc7d310ed6044ba4fc5ef91a6c382d1c9a6
Found 15 .pgn(.gz) files, creating 15 chunks for processing.
```

### DIFF
--- a/scoreWDLstat.cpp
+++ b/scoreWDLstat.cpp
@@ -734,6 +734,11 @@ int main(int argc, char const *argv[]) {
         }
 
         std::cout << "Filtering pgn files with nElo in [" << mi << ", " << ma << "]" << std::endl;
+        if (mi != -ma && !cmd.has_argument("--SPRTonly", true)) {
+            std::cout << "Warning: Asymmetric nElo window suggests --SPRTonly should be used!"
+                      << std::endl;
+        }
+
         filter_files(files_pgn, meta_map, EloFilterStrategy(mi, ma));
     }
 
--- a/scoreWDLstat.cpp
+++ b/scoreWDLstat.cpp
@@ -475,11 +475,11 @@ class EloFilterStrategy {
 
     double pentanomialToEloDiff(const std::vector<int> &pentanomial) const {
         auto pairs            = std::accumulate(pentanomial.begin(), pentanomial.end(), 0);
-        const double WW       = double(pentanomial[0]) / pairs;
-        const double WD       = double(pentanomial[1]) / pairs;
+        const double WW       = double(pentanomial[4]) / pairs;
+        const double WD       = double(pentanomial[3]) / pairs;
         const double WLDD     = double(pentanomial[2]) / pairs;
-        const double LD       = double(pentanomial[3]) / pairs;
-        const double LL       = double(pentanomial[4]) / pairs;
+        const double LD       = double(pentanomial[1]) / pairs;
+        const double LL       = double(pentanomial[0]) / pairs;
         const double score    = WW + 0.75 * WD + 0.5 * WLDD + 0.25 * LD;
         const double WW_dev   = WW * std::pow((1 - score), 2);
         const double WD_dev   = WD * std::pow((0.75 - score), 2);
--- a/scoreWDLstat.cpp
+++ b/scoreWDLstat.cpp
@@ -466,6 +466,47 @@ class ThreadsFilterStrategy {
     }
 };
 
+class EloFilterStrategy {
+    double EloDiff;
+
+   public:
+    EloFilterStrategy(double ed) : EloDiff(ed) {}
+
+    double pentanomialToEloDiff(const std::vector<int> &pentanomial) const {
+        auto pairs            = std::accumulate(pentanomial.begin(), pentanomial.end(), 0);
+        const double WW       = double(pentanomial[0]) / pairs;
+        const double WD       = double(pentanomial[1]) / pairs;
+        const double WLDD     = double(pentanomial[2]) / pairs;
+        const double LD       = double(pentanomial[3]) / pairs;
+        const double LL       = double(pentanomial[4]) / pairs;
+        const double score    = WW + 0.75 * WD + 0.5 * WLDD + 0.25 * LD;
+        const double WW_dev   = WW * std::pow((1 - score), 2);
+        const double WD_dev   = WD * std::pow((0.75 - score), 2);
+        const double WLDD_dev = WLDD * std::pow((0.5 - score), 2);
+        const double LD_dev   = LD * std::pow((0.25 - score), 2);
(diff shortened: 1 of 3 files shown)


scoreWDLstat.cpp
```diff
@@ -466,6 +466,47 @@ class ThreadsFilterStrategy {
...
 };
 
+class EloFilterStrategy {
+    double MaxEloDiff;
+
+   public:
+    EloFilterStrategy(double m) : MaxEloDiff(m) {}
+
+    double pentanomialToEloDiff(const std::vector<int> &pentanomial) const {
+        auto pairs            = std::accumulate(pentanomial.begin(), pentanomial.end(), 0);
```
Disservin, line 476: can you add the `<numeric>` header as well please?

robertnurnberg:
One thought I had: now that we filter for Elo difference in the update script, we could drop `--SPRTonly`. What do you think? @vondele 
Could either do it here or in a follow-up PR.

scoreWDLstat.cpp
```diff
@@ -467,10 +467,10 @@ class ThreadsFilterStrategy {
 };
 
 class EloFilterStrategy {
-    double EloDiff;
+    double MaxEloDiff;
```
Disservin, line 470: well if you want to make the names consistent, i think all variables are in "snake_case" :D

robertnurnberg: T'was more about output and parameter name. For the names I would stick to the shadowing of the cli parameter for now. But I would not mind if you change the names in an upcoming clean-up. ;)

vondele:
> One thought I had: now that we filter for Elo difference in the update script, we could drop `--SPRTonly`. What do you think? @vondele Could either do it here or in a follow-up PR.

I don't think we should drop the option from the code (might be useful later on), but we could drop from our default way of creating the model. I see you didn't change yet how we create the model, I assume you want to do some experimentation first.

vondele:
One thing that might be interesting for experimentation would be to have an Elo window for tests to include, so we could specifically look at how the model changes with Elo advantage (e.g. --EloDiffLowerBound --EloDiffHigherBound).

robertnurnberg:
So I made the penta order consistent with fishtest (I believe) and dropped SPRTonly.

Regarding the window: Possible of course. One thing to keep in mind is that then we would have to exclude PTs again, right? Because there master is `new`, so penta stats will be inverted?

Let me know how to proceed: (a) keep abs diff as in PR, (b) change PR to allow for window.

robertnurnberg:
I guess one way forward is to allow the window in the cpp code, but for the update script only use symmetric windows. Then the user needs to be careful (and use `--SPRTonly` for asymmetric windows fed to the cpp code).

robertnurnberg:
Pushed changes to Elo difference window. The update script will always use a symmetric window.
```diff
> diff scoreWDLstat_master.log scoreWDLstat.log 
7,10c7,11
< Found 13 .pgn(.gz) files, creating 13 chunks for processing.
Progress: 13/13
< Time taken: 30.378s
< Wrote 64384927 scored positions from 960512 games to updateWDL.json for analysis.
---
> Filtering pgn files with nElo in [-5, 5]
> Found 12 .pgn(.gz) files, creating 12 chunks for processing.
Progress: 12/12
> Time taken: 27.413s
> Wrote 62965876 scored positions from 939616 games to updateWDL.json for analysis.
```

PR #194, opened 2024-05-30
reflect the WDL model update in Stockfish

robertnurnberg:
Draft for now, as it still needs the change to `default_firstrev` in `updateWDL.sh`.


robertnurnberg:
Ready for merge now.

PR #198, opened 2024-07-03
Use np.nan instead of np.NaN

### LINKED ISSUE #197: scoreWDL.py Breaks with Numpy 2.0 due to usage of np.NaN
```
  File "/Users/shawn/WDL_model/scoreWDL.py", line 729, in <module>
    wdl_data.load_json_data(args.filename)
  File "/Users/shawn/WDL_model/scoreWDL.py", line 150, in load_json_data
    self.w_density = np.full_like(total, np.NaN, dtype=float)
                                         ^^^^^^
  File "/Users/shawn/WDL_model/venv/lib/python3.12/site-packages/numpy/__init__.py", line 397, in __getattr__
    raise AttributeError(
AttributeError: `np.NaN` was removed in the NumPy 2.0 release. Use `np.nan` instead.. Did you mean: 'nan'?
```

xu-shawn:
Fixes #197

### DIFF
--- a/scoreWDL.py
+++ b/scoreWDL.py
@@ -147,9 +147,9 @@ class WdlData:
         # define wdl densities: if total == 0, entries will be NaN (useful for contour plots)
         total = self.wins + self.draws + self.losses
         self.mask = total > 0
-        self.w_density = np.full_like(total, np.NaN, dtype=float)
-        self.d_density = np.full_like(total, np.NaN, dtype=float)
-        self.l_density = np.full_like(total, np.NaN, dtype=float)
+        self.w_density = np.full_like(total, np.nan, dtype=float)
+        self.d_density = np.full_like(total, np.nan, dtype=float)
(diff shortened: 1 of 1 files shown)


robertnurnberg:
LGTM

Disservin:
mh dunno if we want to now require a version in https://github.com/official-stockfish/WDL_model/blob/master/requirements.txt ?

Disservin:
test

xu-shawn:
> mh dunno if we want to now require a version in https://github.com/official-stockfish/WDL_model/blob/master/requirements.txt ?

This shouldn't require any dependency changes. AFAIK `np.nan` has existed since numpy 1.0

robertnurnberg:
Agree with Shawn. Prior to Numpy 2.0 `np.NaN` (in fact any way to capitalize it) was an accepted alias for `np.nan`.

PR #201, opened 2025-01-21
Improve stof

Disservin:
Use the more correct and fast implementation from the STL to parse floating point numbers.
Upcoming clang release finally implements it and gcc has had this for quite some time now, since gcc 11.1.

Also do some proper pointer checks for the custom implementation.

No functional change


robertnurnberg:
Non functional in my local test with gcc 11.4.

PR #202, opened 2025-01-22
update chess-lib to 0.7.3

robertnurnberg:
This incorporates a recent fix that stops reading PGN's at the first ambiguous move.

### DIFF
--- a/external/chess.hpp
+++ b/external/chess.hpp
@@ -25,7 +25,7 @@ THIS FILE IS AUTO GENERATED DO NOT CHANGE MANUALLY.
 
 Source: https://github.com/Disservin/chess-library
 
-VERSION: 0.6.56
+VERSION: 0.7.3
 */
 
 #ifndef CHESS_HPP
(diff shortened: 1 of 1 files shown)


Disservin:
ideally there's a try and catch around the san for AmbiguousMoveError and then you skip the pgn, so that the other pgns in the same file are still read

robertnurnberg:
Oh, I thought the pgn reader would just skip the game. Is that possible to do on the chess-lib side?

Disservin:
well 
a) an exception is raised, which is caught too far away so the reader stops, this is the same in python-chess
b) dunno about enabling some sort of auto skip, not really visible for a user perspective

robertnurnberg:
agreed. see my edit above.

robertnurnberg:
PS: While at it, we would also check for `NO_MOVE` return from the `parseSAN` function. As in:

```c++
    m = uci::parseSan(board, move, moves);

    // chess-lib may call move() with empty strings for move
    if (m == Move::NO_MOVE) {
      this->skipPgn(true);
      return;
    }

    board.makeMove<true>(m);
```

PR #211, opened 2025-03-13
Update chess-library to 0.8.4

Disservin:
Uses error codes during parsing instead of exceptions. Through fuzzing a few minor issues were fixed, which could cause a parser crash, which shouldn't be the case anymore.

### DIFF
--- a/external/chess.hpp
+++ b/external/chess.hpp
@@ -25,7 +25,7 @@ THIS FILE IS AUTO GENERATED DO NOT CHANGE MANUALLY.
 
 Source: https://github.com/Disservin/chess-library
 
-VERSION: 0.7.4
+VERSION: 0.8.4
 */
 
 #ifndef CHESS_HPP
@@ -75,6 +75,22 @@ class Color {
             color = underlying::WHITE;
         } else if (str == "b") {
             color = underlying::BLACK;
+        }
+    }
+
+    /**
+     * @brief Gets the long string representation of the color
+     * @return "White" for WHITE, "Black" for BLACK
+     * "None" for NONE
+     */
+    [[nodiscard]] std::string longStr() const {
+        switch (color) {
+            case underlying::WHITE:
+                return "White";
+            case underlying::BLACK:
+                return "Black";
+            default:
+                return "None";
         }
     }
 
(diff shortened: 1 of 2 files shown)


robertnurnberg:
@vondele did you test this locally? for me this is functional when running with

```
./updateWDL.sh --firstrev f3bfce353168b03e4fedce515de1898c691f81ec
```

```diff
< Progress: 16/16
< Time taken: 20.304s
< Wrote 137192603 scored positions from 2057929 games to updateWDL.json for analysis.
---
> Progress: 15/16
> Time taken: 17.205s
> Wrote 125762555 scored positions from 1886610 games to updateWDL.json for analysis.
```

So patchs finds fewer positions and reads fewer games. @Disservin should we use the new lib differently, to still get the discared games?

Disservin:
Yikes. There shouldn’t be a functional difference I believe and it was non functional when I tested it.. can you figure out the pgns which deviate? Then I can take a look at what’s changed.

there’s no error reported right ?

robertnurnberg:
No errors, no. Will be busy today, can look tonight. The command `./updateWDL.sh --firstrev f3bfce353168b03e4fedce515de1898c691f81ec`  only looks at 16 pgn in total, so shouldn't be too hard to find.

Disservin:
The progress is the chunk progress, each chunk has multiple pgns, which also means that 15/16 is a weird final output because it means that a chunk didn't finish.

robertnurnberg:
In this case chunk number is equal to pgn file number. I checked the logs. You could try the last few LTC tests and see if scoreWDLstat is functional.

Disservin:
i did download all of the recent tests using the download script but everything was the same

robertnurnberg:
Hm, did you also use the update script with the above parameters? That uses Elo info for example...

Disservin:
> Hm, did you also use the update script with the above parameters?

Just did but still the same output

Disservin:
okay I think i found a bug, when you have the time you can retry with this version https://raw.githubusercontent.com/Disservin/chess-library/6f1c9dbaadb79a9881b7c7f1ec90ddfb9ce89454/include/chess.hpp

robertnurnberg:
Good job. The update script now runs identically to old master.

For a much larger analysis I get only a change in error message, and the game number differs by one. Not sure if you would like to compare of parsing of that particular pgn ....

```diff
< Progress: 30/87Error when parsing: ./pgns/25-01-19/678cd339f4dc0a8b4ae8ddaa/678cd339f4dc0a8b4ae8ddaa.pgn.gz
< Failed to parse san. At step 3: ch r1bqkbnr/1p2npp1/p1pp3p/3Pp3/2P1P3/2N2N2/PP2BPPP/R1BQK2R b KQkq - 1 8
---
> Progress: 30/87Error while parsing: ./pgns/25-01-19/678cd339f4dc0a8b4ae8ddaa/678cd339f4dc0a8b4ae8ddaa.pgn.gz. Error: Invalid header: missing closing bracket

< Time taken: 779.435s
< Wrote 10780479686 scored positions from 79174042 games to scoreWDLstat.json for analysis.
---
> Time taken: 876.074s
> Wrote 10780479686 scored positions from 79174041 games to scoreWDLstat.json for analysis.
```

vondele:
ok, thanks for looking into this, I didn't run it locally....

PR #212, opened 2025-03-14
fix pgn parsing bug

Disservin:


robertnurnberg:
LGTM

No functional change compared to previous master.

